PR #348, opened 2024-04-30
Tooltips in copilot

SanderGi:
Task: https://app.asana.com/0/1203067047205953/1206050253347928

### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```


anish-work:
NEW PR - https://github.com/GooeyAI/gooey-server/pull/443
Closing this!

PR #350, opened 2024-05-02
Animation ux

clr-li:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```


gooey_ui/components/__init__.py
```diff
@@ -690,6 +690,62 @@ def horizontal_radio(
...
+        key=key,
+        help=help,
+        disabled=disabled,
+        label_visibility=label_visibility,
+        validate_value=False,
+    )
+    is_custom = value not in options
+    custom = {
+        "label": "Custom",
+        "input": lambda label, key: number_input(
```
devxpy, line 718: use of lambda discouraged

recipes/DeforumSD.py
```diff
@@ -26,9 +27,13 @@ class AnimationModels(TextChoices):
 class _AnimationPrompt(TypedDict):
     frame: str
     prompt: str
+    second: float
 
 
 AnimationPrompts = list[_AnimationPrompt]
+ZoomSettings: dict[int, float] = {0: 1.004}
```
devxpy, line 34: Can't have global variables that are shared across threads

recipes/DeforumSD.py
```diff
@@ -220,32 +353,49 @@ def related_workflows(self) -> list:
...
+        with col2:
+            st.write("*End of Video*")
+
+        st.write("#### Step 2: Increase Animation Quality")
+        st.write(
+            "Once you like your keyframes, increase your frames per second for high quality"
+        )
+        st.custom_radio(
+            """###### FPS (Frames per second)""",
+            options=[2, 10, 24],
```
devxpy, line 382: needs refactor, values repeated

PR #389, opened 2024-06-18
Fix Integration Connection Bugs

SanderGi:
Task: https://app.asana.com/0/1203067047205953/1207473549199484

### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```


routers/slack_api.py
```diff
@@ -140,12 +157,12 @@ def slack_connect_redirect(request: Request):
             BotIntegration.objects.filter(pk=bi.pk).update(**config)
             bi.refresh_from_db()
 
+    redirect = connect(bi, current_run, published_run)
```
devxpy: This broke the creation of personal channels as it always set slack_create_personal_channels=False

routers/facebook_api.py
```diff
@@ -146,15 +153,31 @@ def fb_connect_redirect(request: Request):
             status_code=400,
         )
 
+    # only connect to pages that are not already connected
```
devxpy, line 156: Any reason for doing this?

devxpy: reverting for now as it seems to keep pages that are disconnected from facebook around as integrations

devxpy: Ok yeah dont wanna loose the analytics for those integrations, fixed in https://github.com/GooeyAI/gooey-server/commit/c08a3c6ad00d7445804b456356515a6edf906195

routers/slack_api.py
SanderGi: I believe sean requested the creation of personal channels disabled

SanderGi: image

SanderGi: I did go through and test all the integrations. Would be good to have some sort of automated test for this too, even if it mocks the 3rd party portions

SanderGi: Tracking here: https://app.asana.com/0/1203067047205953/1204911106314617

devxpy: I am not sure how we enable creation of personal channels in this case. There would be no way for end users to do it, so maybe we need some ux around this too

PR #405, opened 2024-07-15
Json format on doc_search, doc_summary, and google_gpt

SanderGi:
Task: https://app.asana.com/0/1203067047205953/1207720614246546
Only for gpt4o, gpt4 turbo vision, and gpt4 turbo

### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/language_model.py
+++ b/daras_ai_v2/language_model.py
@@ -33,6 +33,10 @@ from daras_ai_v2.text_splitter import (
 from functions.recipe_functions import LLMTools
 
 DEFAULT_SYSTEM_MSG = "You are an intelligent AI assistant. Follow the instructions as closely as possible."
+DEFAULT_JSON_PROMPT = (
+    "Please respond directly in JSON format. "
+    "Don't output markdown or HTML, instead print the JSON object directly without formatting."
+)
 
 CHATML_ROLE_SYSTEM = "system"
 CHATML_ROLE_ASSISTANT = "assistant"
@@ -143,6 +147,7 @@ class LargeLanguageModels(Enum):
         llm_api=LLMApis.openai,
         context_window=4096,
         price=1,
+        supports_json=True,
     )
     gpt_3_5_turbo_16k = LLMSpec(
         label="ChatGPT 16k (openai)",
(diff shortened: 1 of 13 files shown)


daras_ai_v2/language_model.py
```diff
@@ -143,6 +147,7 @@ class LargeLanguageModels(Enum):
         llm_api=LLMApis.openai,
         context_window=4096,
         price=1,
+        supports_json=True,
```
SanderGi, line 151: Have you tested this? When I went through and tested all the open ai models, this one did not work for me 🤔

daras_ai_v2/language_model.py
```diff
@@ -208,20 +213,23 @@ class LargeLanguageModels(Enum):
...
     )
     gemma_7b_it = LLMSpec(
         label="Gemma 7B (Google)",
         model_id="gemma-7b-it",
         llm_api=LLMApis.groq,
         context_window=8_192,
         price=1,
+        supports_json=True,
```
SanderGi, line 232: Shouldn’t this include validation and retries to make sure the output is JSON similar to the openai implementation?

devxpy: groq supports json mode

devxpy: https://console.groq.com/docs/text-chat#json-mode-object-object

daras_ai_v2/language_model.py
devxpy: I just copied the ones listed here - https://platform.openai.com/docs/guides/json-mode

devxpy: in any case it should work for all models to some degree because of the added json prompt

SanderGi: Yes, but when I run gpt 3.5 with the json flag enabled, it errors out in practice: https://gooey.ai/compare-large-language-models/?run_id=oygoftsplu02&uid=2pZJQouaP2Okm18cbEhIpOQpqCg1

devxpy: i guess not though, it might lead to a misuse of json mode on non json mode models

SanderGi: The supports_json flag above was meant to be used to disable json mode for models that don't support it. I'll leave it up to you whether we prefer showing the JSON format ui but throwing an error in that case (or perhaps even implement our own JSON mode that retries a couple of times if the response happens to be invalid json) over ignoring it for unsupported models

devxpy: What we have now is ok. 

PS doing json mode well requires constrained decoding https://lmsys.org/blog/2024-02-05-compressed-fsm/

PR #414, opened 2024-07-19
Integration settings for twilio

SanderGi:
Task: https://app.asana.com/0/1203067047205953/1207805759212356
No connect UI, just the integration settings as requested.

### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


devxpy:
Just a thought - can we remove twilio tts setting from bot integration, given that we already list the google voices in the saved run? If we already support google tts, we can just change the logic to say if the saved run has selected google tts, we use twilio inbuilt google tts with the same voice. 
we can add support for polly too in time but not needed for now.

similarly for the asr model, if its set to auto detect in the saved run, we can pick up the language setting and use that for twilio too


I have seen in user testing that its hard to have 2 places to configure things + this makes it slightly easier to test without releasing the bot integration

this looks pretty good to me otherwise

SanderGi:
> Just a thought - can we remove twilio tts setting from bot integration, given that we already list the google voices in the saved run? If we already support google tts, we can just change the logic to say if the saved run has selected google tts, we use twilio inbuilt google tts with the same voice. we can add support for polly too in time but not needed for now.
> 
> similarly for the asr model, if its set to auto detect in the saved run, we can pick up the language setting and use that for twilio too
> 
> I have seen in user testing that its hard to have 2 places to configure things + this makes it slightly easier to test without releasing the bot integration
> 
> this looks pretty good to me otherwise

What about the current use case for these settings where asr/tts isn’t enabled on the underlying gooey run? Should we just default to an arbitrary twilio tts voice and english for asr?

devxpy:
> > Just a thought - can we remove twilio tts setting from bot integration, given that we already list the google voices in the saved run? If we already support google tts, we can just change the logic to say if the saved run has selected google tts, we use twilio inbuilt google tts with the same voice. we can add support for polly too in time but not needed for now.
> > similarly for the asr model, if its set to auto detect in the saved run, we can pick up the language setting and use that for twilio too
> > I have seen in user testing that its hard to have 2 places to configure things + this makes it slightly easier to test without releasing the bot integration
> > this looks pretty good to me otherwise
> 
> What about the current use case for these settings where asr/tts isn’t enabled on the underlying gooey run? Should we just default to an arbitrary twilio tts voice and english for asr?

Yup, the default we hardcoded now can work. We can also force people to do this at time of creation of the twilio bot, but probably ok to use the hardcoded defaults right now

devxpy:
I just realised the user language support list can be different from the Twilio asr supported languages. 

If such an edge case happens, just fallback to gooey asr

daras_ai_v2/asr.py
```diff
@@ -440,6 +429,30 @@ def asr_language_selector(
     )
 
 
+def get_language_match(lang: str | None, languages: list[str]) -> str | None:
```
devxpy, line 432: why not use `normalised_lang_in_collection` ??

routers/twilio_api.py
```diff
@@ -41,6 +43,42 @@
...
 
+def get_twilio_tts_voice(bi: BotIntegration) -> str:
+    from recipes.TextToSpeech import TextToSpeechProviders
+
+    run = bi.get_active_saved_run()
+    state: dict = run.state
+    tts_provider = state.get("tts_provider")
+    if tts_provider and tts_provider == TextToSpeechProviders.GOOGLE_TTS.name:
+        voice = "Google." + state.get("google_voice_name", "en-US-Wavenet-F")
+        if voice not in TWILIO_SUPPORTED_VOICES:
```
devxpy, line 54: why is use_gooey_tts True for this case?

daras_ai_v2/asr.py
SanderGi: My bad. Didn’t realize we had a function for this. I looked at the asr code and only saw the place were we manually did it inline without calling a function so thought we didn’t have this refactored yet

routers/twilio_api.py
SanderGi: Is it? I have the following python in there that shouldn’t set it to true? It seems to work for me, I’ll double check

self.use_gooey_tts = (
            tts_provider and tts_provider != TextToSpeechProviders.GOOGLE_TTS.name
        )

daras_ai_v2/asr.py
SanderGi: [01e3c81](https://github.com/GooeyAI/gooey-server/pull/414/commits/01e3c81a1c3afe488e24192ca928ddc5728a289b)

routers/twilio_api.py
SanderGi: [e567511](https://github.com/GooeyAI/gooey-server/pull/414/commits/e56751148d698fcfc89d117e23fc1bf1ebe4b113)

devxpy:
image
??

SanderGi:
> image ??

Is this before or after your refactor? You probably just set the phone number sid incorrectly in the db

PR #421, opened 2024-08-01
Store Stripe billing info + auto recharge without subscription

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


devxpy:
can you review this once please

app_users/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+        pm.customer,
+        invoice_settings=dict(default_payment_method=pm),
+    )
+
+    # if user doesn't already have a active billing/autorecharge info, so we don't need to do anything
+    # set user's subscription to the free plan
+    if (
+        reason == TransactionReason.ADDON
+        and not Subscription.objects.filter(
+            user__uid=uid, payment_provider__isnull=False
```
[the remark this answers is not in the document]
devxpy: This is checking if the user has payment info stored or not. When the user unsubscribes from paypal or deletes all payment info, we clear out the payment provider

daras_ai_v2/billing.py
```diff
@@ -414,57 +382,66 @@ def render_paypal_addon_buttons():
 
 
 def render_stripe_addon_buttons(user: AppUser):
+    if not (user.subscription and user.subscription.payment_provider):
+        save_pm = gui.checkbox(
+            "Save payment method for future purchases & auto-recharge", value=True
+        )
+    else:
+        save_pm = True
```
nikochiko, line 390: isn't there a checkbox in the next step on stripe's page anyways?

devxpy: Stripe shows different payment options depending on this checkbox!

daras_ai_v2/billing.py
```diff
@@ -473,55 +450,98 @@ def stripe_addon_checkout_redirect(user: AppUser, dollat_amt: int):
...
+            success_url=get_app_route_url(payment_processing_route),
+            cancel_url=get_app_route_url(account_route),
+            allow_promotion_codes=True,
+            customer=customer,
+            line_items=line_items,
+            metadata=metadata,
+            subscription_data={"metadata": metadata},
+            saved_payment_method_options={"payment_method_save": "enabled"},
+        )
+        raise gui.RedirectException(checkout_session.url, status_code=303)
```
nikochiko, line 544: :+1:

app_users/tasks.py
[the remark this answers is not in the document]
devxpy: xref 
IMG_7702

daras_ai_v2/billing.py
devxpy: The one on stripe is also easy to miss and unchecked by default

daras_ai_v2/billing.py
```diff
@@ -544,44 +564,90 @@ def render_paypal_subscription_button(
...
-        if gui.button(f"{icons.edit} Edit", type="link", key="manage-payment-provider"):
-            raise gui.RedirectException(user.subscription.get_external_management_url())
+    if not user.subscription:
+        return
 
     pm_summary = gui.run_in_thread(
         user.subscription.get_payment_method_summary, cache=True
     )
     if not pm_summary:
         return
```
nikochiko, line 574: `pm_summary` is often None, and takes a while to load.

When that happens, the cancel button won't be shown.

devxpy: Ok - show this button based on where the user’s subscription has a payment_provider or not?

devxpy: Actually I think this is good. It handles cases where the payment method is saved in stripe/paypal but not in our db?

devxpy: See https://github.com/GooeyAI/gooey-server/blob/9c52e1410c4bb725c49ee272440bbe62192058e5/payments/models.py#L200

nikochiko:
thanks! this looks great when I tried it out.

devxpy:
lgtm

app_users/tasks.py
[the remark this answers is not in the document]
nikochiko: There could be a bug when:
1. save a payment method with addon payment
2. remove the payment method from stripe
3. auto recharge is still enabled but there's no payment method that user sees in /account because payment_provider is None

i think it's not a problem though because auto recharge payments would fail (no payment method). so it's OK.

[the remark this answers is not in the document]
devxpy: [https://github.com/GooeyAI/gooey-server/blob/2c8ab553ac6b665ab201a20c088cf0fa29923729/payments/auto_recharge.py#L111](https://github.com/GooeyAI/gooey-server/blob/2c8ab553ac6b665ab201a20c088cf0fa29923729/payments/auto_recharge.py#L111) :)

PR #422, opened 2024-08-01
OpenAPI revisions for compatibility with Fern SDK generator

nikochiko:
# Change Summary
* ~~auth~~
  * ~~new `APIAuth` class that inherits from `fastapi.security.base.SecurityBase` (refactored into this from `api_auth_header` to output OpenAPI with bearer scheme)~~
  * ~~rename `auth_keyword` to `auth_scheme`~~
* method names with fern-specific openapi attributes
  * add `get_openapi_extra` classmethod to `BasePage`
  * add `sdk_method_name` class attribute to `BasePage`
  * `x-fern-ignore` API endpoints that shouldn't be present in SDK
* Model classes
  * Rename `Text2ImgModels` to `TextToImageModels` for parity with SDK names
  * Rename `Img2ImgModels` to `ImageToImageModels` for parity with SDK names
  * Rename `LipsyncModel` to `LipsyncModels` (plural) for consistency with rest of model classes 
* `GooeyEnum`
  * add `api_enum` class property for `GooeyEnum` that creates a dynamic enum mapping from `api_value`->`api_value` (default, `name`->`name`)
  * remove `api_choices` in favor of `api_enum`
  * make use of `GooeyEnum` for
    * `AnimationModels`
    * `AsrModels`
    * `TextToImageModels`
    * `ImageToImageModels`
    * `SegmentationModels`
    * `UpscalerModels`
    * `ControlNetModels`
    * `AsrOutputFormat`
    * `CitationStyles`
    * `Schedulers`
    * `TextToSpeechProviders`
    * `TextToAudioModels`
    * `CombineDocumentChainType`
    * `SerpSearchType`
    * `SerpSearchLocations`

### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

---
  - https://app.asana.com/0/0/1208315549936081
  - https://app.asana.com/0/0/1207921877486111


devxpy:
do we really need this huge refactor of all enums?

devxpy:
in any case if this can be broken up between the auth header and the enum that would be greatly appreciated

nikochiko:
> do we really need this huge refactor of all enums?

I used `GooeyEnum` where we were already using Enums or TextChoices per https://github.com/GooeyAI/gooey-server/discussions/393. That was an easier change and consistent with this discussion.

Where it was something else, I have used `x-fern-type-name` to not make new Enums.

devxpy:
> > do we really need this huge refactor of all enums?
> 
> I used `GooeyEnum` where we were already using Enums or TextChoices per https://github.com/GooeyAI/gooey-server/discussions/393. That was an easier change and consistent with this discussion.
> 
> Where it was something else, I have used `x-fern-type-name` to not make new Enums.


Do we need the enum for fern to generate the `literal | str` type annotation? The enum refactor is nice but i think it would need a lot of testing to pass the review

nikochiko:
By default, Fern will generate types like

```python
CompareLLMSelectedModelsItem = typing.Union[typing.Literal[...], typing.Any]
```

Giving an Enum type lets us customize this name because fastapi will generate an OpenAPI component with the name inferred from the Enum's name.

By default, Literals will create a new OpenAPI Enum type for each input, and each will have its own different name.

nikochiko:
Separated into this and https://github.com/GooeyAI/gooey-server/pull/465

nikochiko:
I have tested all the affected recipes in this PR - the UX names, API names, and their use in `run` & `run_v2`.

The nice thing with the refactor is that it doesn't affect how the value is stored in state, or passed down to the `run` method of each recipe. The automated tests already guarantee that all pages load with 200, `sane_defaults` work as before. We can continue testing like before manually after deployment, and also revert easily in case anything is breaking (because no migrations).

PR #453, opened 2024-08-30
fix: ux behavior on save for anonymous users

nikochiko:
should redirect to login page with next=...

### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

---
  - https://app.asana.com/0/0/1208189770838403


devxpy:
curious if this is upto date as per 
> @SeanBlagsvedt This is the intended UX flow for anonymous users.

daras_ai_v2/base.py
```diff
@@ -514,23 +520,35 @@ def _render_published_run_save_buttons(
...
             if publish_modal.is_open():
                 with publish_modal.container(style={"minWidth": "min(500px, 100vw)"}):
                     self._render_publish_modal(
                         current_run=current_run,
                         published_run=published_run,
                         modal=publish_modal,
-                        is_update_mode=can_edit,
                     )
 
+    def _save_for_anonymous_user(self, current_run: SavedRun):
```
devxpy, line 535: believe you dont need to pass current_run around anymore?

daras_ai_v2/base.py
```diff
@@ -514,23 +520,35 @@ def _render_published_run_save_buttons(
...
+        if self._has_request_changed():
```
devxpy, line 536: I believe what @SeanBlagsvedt asked for is that the run not be actually charged under anon but under the logged in user?

daras_ai_v2/base.py
```diff
@@ -514,23 +520,35 @@ def _render_published_run_save_buttons(
...
+            current_run = self.on_submit()
+            if not current_run:
+                gui.rerun()
+
+        self.ensure_authentication(
```
devxpy, line 541: this should do only redirect not the check for auth given this function is called `_save_for_anonymous_user`

daras_ai_v2/base.py
```diff
@@ -607,46 +624,75 @@ def _render_publish_modal(
 
         self._render_admin_options(current_run, published_run)
 
-        if not pressed_save:
-            return
+        update_existing_published_run = self.can_user_edit_published_run(published_run)
+        if pressed_save:
```
devxpy, line 628: this if statement is better written as `if not <condn>: return`

daras_ai_v2/base.py
```diff
@@ -607,46 +624,75 @@ def _render_publish_modal(
 
         self._render_admin_options(current_run, published_run)
 
-        if not pressed_save:
-            return
+        update_existing_published_run = self.can_user_edit_published_run(published_run)
```
devxpy, line 627: only need to check this after pressed_save condition

daras_ai_v2/base.py
```diff
@@ -607,46 +624,75 @@ def _render_publish_modal(
...
+        if published_run:
+            # update existing published run
             updates = dict(
                 saved_run=current_run,
-                title=published_run_title.strip(),
-                notes=published_run_notes.strip(),
-                visibility=published_run_visibility,
+                title=title or published_run.title,
+                notes=(
+                    notes if notes is not None else published_run.notes
```
devxpy, line 669: inline if condition discouraged

daras_ai_v2/base.py
```diff
@@ -607,46 +624,75 @@ def _render_publish_modal(
...
-                return
+            self.save_or_update_published_run_and_redirect(
+                current_run=current_run,
+                published_run=(
+                    published_run if update_existing_published_run else None
+                ),
+                title=published_run_title.strip(),
+                notes=published_run_notes.strip(),
+                visibility=published_run_visibility,
+            )
```
devxpy, line 647: refactor as `if ... else`

daras_ai_v2/base.py
```diff
@@ -607,46 +624,75 @@ def _render_publish_modal(
...
 
-        if not pressed_save:
-            return
+        update_existing_published_run = self.can_user_edit_published_run(published_run)
+        if pressed_save:
+            if not (published_run.is_root() and update_existing_published_run):
+                # don't validate title root example
+                try:
+                    self._validate_published_run_title(
+                        gui.session_state["published_run_title"]
```
devxpy, line 633: use local var `published_run_title`

nikochiko:
> curious if this is upto date as per
> 
> > @SeanBlagsvedt This is the intended UX flow for anonymous users. https://docs.google.com/presentation/d/1P5MKF0jykYPw-G_9SgvKEpu7fjD2v-0gZgLJB4Uw174/edit#slide=id.g2ef5e525406_0_329

needs an update, if I remember right

daras_ai_v2/base.py
```diff
@@ -343,6 +345,12 @@ def render(self):
         self._user_disabled_check()
         self._check_if_flagged()
 
+        if self.should_submit_after_login():
```
devxpy, line 348: SAVE_AFTER_LOGIN_Q and SUBMIT_AFTER_LOGIN_Q shold be mutually exclusive

daras_ai_v2/base.py
```diff
@@ -99,6 +87,8 @@
 gooey_rng = Random()
 
 SUBMIT_AFTER_LOGIN_Q = "submitafterlogin"
+SAVE_AFTER_LOGIN_Q = "saveafterlogin"
```
devxpy, line 90: The save terminology is very confusing given we are using published run everywhere else. Use `PUBLISH_AFTER_LOGIN_Q`

daras_ai_v2/base.py
```diff
@@ -1621,6 +1723,13 @@ def should_submit_after_login(self) -> bool:
             and not self.request.user.is_anonymous
         )
 
+    def should_save_after_login(self) -> bool:
```
devxpy, line 1726: should_create_published_run_after_login

daras_ai_v2/base.py
```diff
@@ -620,28 +705,36 @@ def _render_publish_form(
...
             )
+            if not self._has_published_run_changed(published_run=pr, **updates):
+                gui.error("No changes to publish", icon="⚠️")
+                return
+            pr.add_version(user=self.request.user, **updates)
         raise gui.RedirectException(pr.get_app_url())
 
+    def _get_default_pr_title(self):
+        recipe_title = self.get_root_pr().title or self.title
+        if self.request.user and not self.request.user.is_anonymous:
```
devxpy, line 732: I believe this should just not run at all if the user is null/anon?

daras_ai_v2/base.py
```diff
@@ -1594,11 +1672,35 @@ def render_extra_waiting_output(self):
...
         sr = self.on_submit()
         if not sr:
             return
-        raise gui.RedirectException(self.app_url(run_id=sr.run_id, uid=sr.uid))
+        if save:
+            self.save_and_redirect(sr)
+        else:
+            raise gui.RedirectException(self.app_url(run_id=sr.run_id, uid=sr.uid))
+
+    def save_and_redirect(self, sr: SavedRun | None = None) -> typing.NoReturn:
```
devxpy, line 1684: create_published_run_and_redirect

daras_ai_v2/base.py
```diff
@@ -325,6 +315,17 @@ def sentry_event_set_user(self, event, hint):
             }
         return event
 
+    def _load_state_from_query_params(self):
+        request_state = self.request.query_params.get(REQUEST_STATE_Q)
+        if not request_state:
+            return
+
+        try:
+            request_state = base64.urlsafe_b64decode(request_state).decode()
```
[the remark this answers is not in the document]
nikochiko: but for anon users, we won't have their real `uid` until they login.

are you suggesting that we create a run without a `uid`, and when the user comes back with `SUBMIT_AFTER_LOGIN_Q`, we then update the `uid` and run?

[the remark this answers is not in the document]
devxpy: we generate a firebase user as soon as you submit, this should do the same thing

daras_ai_v2/base.py
nikochiko: done

daras_ai_v2/base.py
```diff
@@ -1594,11 +1672,35 @@ def render_extra_waiting_output(self):
...
+            self.save_and_redirect(sr)
+        else:
+            raise gui.RedirectException(self.app_url(run_id=sr.run_id, uid=sr.uid))
+
+    def save_and_redirect(self, sr: SavedRun | None = None) -> typing.NoReturn:
+        pr = self.create_published_run(
+            published_run_id=get_random_doc_id(),
+            saved_run=sr or self.current_sr,
+            user=self.request.user,
+            title=gui.session_state.get(
```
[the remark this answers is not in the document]
nikochiko: fixed

daras_ai_v2/base.py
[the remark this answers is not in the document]
nikochiko: for anon users who come back with an existing account, we would have 2 different uids.

i fixed this case by setting the uid of the unqueued run when user comes back. i added a concurrency check there too so that we don't queue the same run twice.

daras_ai_v2/base.py
```diff
@@ -490,39 +574,49 @@ def _render_published_run_save_buttons(self, *, sr: SavedRun, pr: PublishedRun):
...
 
-            ref = gui.use_confirm_dialog(key="publish-modal", close_on_confirm=False)
+            ref = gui.use_alert_dialog(key="publish-modal")
             if gui.button(
-                f'<i class="fa-regular fa-floppy-disk"></i> <span class="d-none d-lg-inline">{label}</span>',
+                f'{icon} <span class="d-none d-lg-inline">{label}</span>',
                 className="mb-0 ms-lg-2 px-lg-4",
                 type="primary",
             ):
+                if not self.request.user or self.request.user.is_anonymous:
```
devxpy, line 589: please put this in an if..else case |  its not clear from this that publish_for_anonymous_user will end up raising an exception and the stuff below won't run

daras_ai_v2/base.py
```diff
@@ -1621,10 +1706,42 @@ def should_submit_after_login(self) -> bool:
...
+            self.request.query_params.get(PUBLISH_AFTER_LOGIN_Q)
+            and self.request.user
+            and not self.request.user.is_anonymous
+        )
+
+    def create_and_validate_new_run(
+        self,
+        *,
+        enable_rate_limits: bool = False,
+        run_status: str | None = "Starting...",
```
devxpy, line 1720: not used

daras_ai_v2/base.py
nikochiko: i added a type annotation for `_publish_for_anonymous_user() -> NoReturn` for this, but i like this because it's even more explicit

daras_ai_v2/base.py
```diff
@@ -490,39 +574,49 @@ def _render_published_run_save_buttons(self, *, sr: SavedRun, pr: PublishedRun):
...
+                    sr=self.current_sr,
+                    pr=self.current_pr,
+                    dialog=ref,
+                    is_update_mode=can_edit,
                 )
 
+    def _publish_for_anonymous_user(self):
+        query_params = {PUBLISH_AFTER_LOGIN_Q: "1"}
+        if self._has_request_changed():
+            sr = self.create_new_run(enable_rate_limits=True, run_status=None)
```
devxpy, line 612: did you mean create_and_validate_run?

nikochiko: ah yes, thanks. good catch

devxpy:
is this finally on me? I saw 2 commits being pushed after it was assigned to me

nikochiko:
> is this finally on me? I saw 3 commits being pushed after it was assigned to me

later two were fixes from the review

devxpy:
> > is this finally on me? I saw 3 commits being pushed after it was assigned to me
> 
> yes, later two were fixes from the review

sorry i didn't realise the review got out

devxpy:
odd behavior
- click on share
- change visibility
- press copy link
- close dialog
- open dialog again, see that visibility hasn't saved

devxpy:
image

close button should be top right corner aligned like so 

image

devxpy:
dialog buttons should be aligned at the corners with equal spacing on bottom and left/right

image
image

example of proper alignment:

image

devxpy:
inconsistency of display between update and options buttons:

- update button shown on all tabs
- options button shown on run, run as API and integrations only

daras_ai_v2/base.py
```diff
@@ -1164,7 +1217,7 @@ def get_sr_pr_from_query_params(
     def get_sr_from_ids(
         cls,
         run_id: str,
-        uid: str,
+        uid: str | None,
```
devxpy, line 1220: not sure why this annotation has changed, it shouldn't be

nikochiko:
> odd behavior
> 
> * click on share
> * change visibility
> * press copy link
> * close dialog
> * open dialog again, see that visibility hasn't saved

That sounds like a reasonable user flow. I'll make it such that the `Copy Link` button also saves like the other `Save` button?

nikochiko:
image

we have bad defaults for padding/margin of headings, this looks uneven too (without any changes made). i'll fix it by adding some top margin/padding to the heading's container too

nikochiko:
better?

image

image

daras_ai_v2/base.py
```diff
@@ -1594,23 +1647,57 @@ def render_extra_waiting_output(self):
...
         sr = self.on_submit()
         if not sr:
             return
         raise gui.RedirectException(self.app_url(run_id=sr.run_id, uid=sr.uid))
 
+    def publish_and_redirect(self) -> typing.NoReturn | None:
+        assert self.request.user and not self.request.user.is_anonymous
+
+        if (
+            self.get_run_state(gui.session_state) == RecipeRunState.starting
```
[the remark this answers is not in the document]
nikochiko: nice, i like this refactor.

yes, `RecipeRunState.starting` was misleading. I initially created a new state `not_started` and later realised that we had `starting` for that already.

daras_ai_v2/base.py
```diff
@@ -1594,23 +1647,57 @@ def render_extra_waiting_output(self):
...
+                    # run has already been started in another tab
+                    gui.rerun()
+                if (
+                    sr.uid == self.current_sr_user.uid
+                    and self.current_sr_user.is_anonymous
+                ):
+                    # run belonged to an anonymous user who now logged
+                    # in from an existing account now, so we update the uid
+                    sr.uid = self.request.user.uid
+                    sr.save(update_fields=["uid"])
```
devxpy, line 1678: PS didn't need to save here as `dump_state_to_sr` would have done it  anyways

daras_ai_v2/base.py
```diff
@@ -1594,23 +1647,57 @@ def render_extra_waiting_output(self):
...
+    def submit_and_redirect(self) -> typing.NoReturn | None:
         sr = self.on_submit()
         if not sr:
             return
         raise gui.RedirectException(self.app_url(run_id=sr.run_id, uid=sr.uid))
 
+    def publish_and_redirect(self) -> typing.NoReturn | None:
+        assert self.request.user and not self.request.user.is_anonymous
+
+        if (
```
devxpy, line 1659: This transaction stuff is more complex than it needs to be. You can do something like:

```py
matches = SavedRun.objects.filter(
    id=CURRENT_RUN.ID,
    run_status="", error_msg="", run_time=datetime.timedelta()  # filter for pending state
).update(run_status=STARTING_STATE, uid=LOGGED_IN_USER.UID)
if matches > 0:  # no matches, run has probably started elsewhere
    >> send task to queue, create published run & redirect
else:
    >> rerun
```

daras_ai_v2/base.py
nikochiko: i'm aware, but i added this because it isn't obvious (also would not be obvious to fix if we were to later refactor `dump_state_to_sr` to only save the state and not everything else)

daras_ai_v2/base.py
nikochiko: thanks

nikochiko:
> That sounds like a plausible user flow. I'll make it such that the Copy Link button also saves like the other Save button

this would require more work than I had thought. right now it doesn't trigger a form submit - all javascript.

nikochiko:
fixed the rest, other than the Copy Link fix.

i'll need to release gooey-gui and update version for that.

devxpy:
> > That sounds like a plausible user flow. I'll make it such that the Copy Link button also saves like the other Save button
> 
> this would require more work than I had thought. right now it doesn't trigger a form submit - all javascript.

You can execute js on a python button I believe
```
if button_pressed:
   gui.js(<js code here>)
```

daras_ai_v2/copy_to_clipboard_button_widget.py
```diff
@@ -23,20 +22,25 @@ def copy_to_clipboard_button(
...
-    class="btn btn-theme btn-{type} {className}"
-    onclick="copyToClipboard(this);"
-    style="{style}"
-    data-clipboard-text="{escape(value)}">
-    {label}
-</button>
-        """,
+    key: str | None = None,
+) -> bool:
+    key = key or "copy-to-clipboard-btn"
```
devxpy, line 30: lets not make the other copy buttons slower

daras_ai_v2/base.py
```diff
@@ -1594,23 +1654,46 @@ def render_extra_waiting_output(self):
...
         if not sr:
             return
         raise gui.RedirectException(self.app_url(run_id=sr.run_id, uid=sr.uid))
 
+    def publish_and_redirect(self) -> typing.NoReturn | None:
+        assert self.request.user and not self.request.user.is_anonymous
+
+        updated_count = SavedRun.objects.filter(
+            id=self.current_sr.id,
+            uid=self.current_sr_user.uid,
```
devxpy, line 1668: this filter wont work if you login with a different user?

nikochiko: `current_sr_user` is different from the current user (it's taken from `current_sr.uid`)

nikochiko: `current_sr.uid` would be clearer here though, thanks!

daras_ai_v2/copy_to_clipboard_button_widget.py
nikochiko: fixed

nikochiko: added a new method `copy_to_clipboard_button_with_return` (had to use chatgpt for name suggestions)

daras_ai_v2/base.py
devxpy: why does this need the uid filter at all? since its filtering on `id`

devxpy:
I get this error when I try to press the options button as a fresh anonymous user 

```
Traceback (most recent call last):
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/uvicorn/protocols/http/httptools_impl.py", line 404, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/fastapi/applications.py", line 270, in __call__
    await super().__call__(scope, receive, send)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/applications.py", line 124, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/Users/dev/Projects/dara/ddgai/./server.py", line 107, in middleware
    await app(scope, receive, send)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/middleware/sessions.py", line 86, in __call__
    await self.app(scope, receive, send_wrapper)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/middleware/authentication.py", line 48, in __call__
    await self.app(scope, receive, send)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/middleware/cors.py", line 92, in __call__
    await self.simple_response(scope, receive, send, request_headers=headers)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/middleware/cors.py", line 147, in simple_response
    await self.app(scope, receive, send)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 75, in __call__
    raise exc
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 64, in __call__
    await self.app(scope, receive, sender)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/routing.py", line 680, in __call__
    await route.handle(scope, receive, send)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/routing.py", line 275, in handle
    await self.app(scope, receive, send)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/routing.py", line 65, in app
    response = await func(request)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/fastapi/routing.py", line 231, in app
    raw_response = await run_endpoint_function(
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/fastapi/routing.py", line 162, in run_endpoint_function
    return await run_in_threadpool(dependant.call, **values)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/starlette/concurrency.py", line 41, in run_in_threadpool
    return await anyio.to_thread.run_sync(func, *args)
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/anyio/to_thread.py", line 33, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 877, in run_sync_in_worker_thread
    return await future
  File "/Users/dev/.virtualenvs/ddgai-b0692ca2/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 807, in run
    result = context.run(func, *args)
  File "/Users/dev/Projects/dara/gooey-gui/py/gooey_gui/core/renderer.py", line 70, in wrapper
    return renderer(
  File "/Users/dev/Projects/dara/gooey-gui/py/gooey_gui/core/renderer.py", line 117, in renderer
    ret = render()
  File "/Users/dev/Projects/dara/ddgai/routers/root.py", line 615, in recipe_or_handle_or_static
    return render_recipe_page(request, parts[0], RecipeTabs.run, example_id)
  File "/Users/dev/Projects/dara/ddgai/routers/root.py", line 691, in render_recipe_page
    page.render()
  File "/Users/dev/Projects/dara/ddgai/./daras_ai_v2/base.py", line 362, in render
    self._render_header()
  File "/Users/dev/Projects/dara/ddgai/./daras_ai_v2/base.py", line 408, in _render_header
    self._render_options_button_with_dialog()
  File "/Users/dev/Projects/dara/ddgai/./daras_ai_v2/base.py", line 473, in _render_options_button_with_dialog
    self._unsaved_options_modal()
  File "/Users/dev/Projects/dara/ddgai/./daras_ai_v2/base.py", line 844, in _unsaved_options_modal
    assert self.request.user
AssertionError
```

daras_ai_v2/base.py
```diff
@@ -1594,23 +1661,46 @@ def render_extra_waiting_output(self):
...
+            id=self.current_sr.id,
+            uid=self.current_sr.uid,
+            # filter for RecipeRunState.standby
+            run_status="",
+            error_msg="",
+            run_time=datetime.timedelta(),
+        ).update(run_status=STARTING_STATE, uid=self.request.user.uid)
+        if updated_count >= 1:
+            # updated now
+            self.dump_state_to_sr(self._get_validated_state(), self.current_sr)
```
devxpy, line 1683: not sure why this is needed here?

nikochiko: ah, good catch. i think it is leftover from the earlier hack where the state was carried over without creating a saved run. thanks!

nikochiko:
@devxpy fixed both in https://github.com/GooeyAI/gooey-server/pull/481

devxpy:
I can still repro this in prod:

 - go to /copilot
 - make an edit to prompt
 - press save and run
 - login
 - run stuck on starting

nikochiko:
fixed in https://github.com/GooeyAI/gooey-server/pull/482 with `self.current_sr.refresh_from_db()` before queueing.

cause was that we updated `current_sr.uid` in the database with `SavedRun.objects.update()`, but not in the ORM object. we should probably get rid of `uid` from the runner task too. `SavedRun.id` should be good enough

devxpy:
this also broke rate limits fixed in https://github.com/GooeyAI/gooey-server/commit/7fe01cb4965cfe2dc5da33c7c015a13be6945231

PR #459, opened 2024-09-09
add GhanaNLP ASR v2 API

anish-work:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

---
  - https://app.asana.com/0/0/1205803182949889
  - https://app.asana.com/0/0/1208264683882283

### DIFF
--- a/daras_ai_v2/asr.py
+++ b/daras_ai_v2/asr.py
@@ -202,6 +202,7 @@ MMS_SUPPORTED = {
 GHANA_NLP_SUPPORTED = {'en': 'English', 'tw': 'Twi', 'gaa': 'Ga', 'ee': 'Ewe', 'fat': 'Fante', 'dag': 'Dagbani',
                        'gur': 'Gurene', 'yo': 'Yoruba', 'ki': 'Kikuyu', 'luo': 'Luo', 'mer': 'Kimeru'}  # fmt: skip
 GHANA_NLP_MAXLEN = 500
+GHANA_API_AUTH_HEADERS = {"Ocp-Apim-Subscription-Key": str(settings.GHANA_NLP_SUBKEY)}
 GHANA_NLP_ASR_V2_SUPPORTED = {
     "tw": "Twi",
     "gaa": "Ga",
@@ -238,6 +239,9 @@ class AsrModels(Enum):
             self.seamless_m4t_v2,
             self.ghana_nlp_asr_v2,
         }
+
+    def expects_wav_data(self) -> bool:
+        return self in {self.ghana_nlp_asr_v2}
 
     @classmethod
     def _deprecated(cls):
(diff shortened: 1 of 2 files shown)


daras_ai_v2/asr.py
```diff
@@ -889,6 +900,18 @@ def run_asr(
             ),
             # queue_prefix="gooey-gpu/short" if is_short else "gooey-gpu/long",
         )
+    elif selected_model == AsrModels.ghana_nlp_asr_v2:
+        r = requests.post(
+            f"https://translation-api.ghananlp.org/asr/v2/transcribe?language={language}",
```
devxpy, line 905: use furl to build urls

daras_ai_v2/asr.py
```diff
@@ -889,6 +900,18 @@ def run_asr(
...
+            headers={
+                "Content-Type": "audio/wav",
+                "Cache-Control": "no-cache",
+                "Ocp-Apim-Subscription-Key": str(settings.GHANA_NLP_SUBKEY),
```
devxpy, line 909: same header variable refactor as in other ghana pr

daras_ai_v2/asr.py
```diff
@@ -889,6 +900,18 @@ def run_asr(
...
+            },
+            data=requests.get(audio_url).content,
```
devxpy, line 911: forgot raise_for_status to this GET request

devxpy: also I think `audio_url_to_wav` will upload the wav file to gcp first and then downloading it here again in same func isn't efficient, have to refactor that

anish-work: @devxpy Updated the code, please review

anish-work:
@dev is this one ready ?

daras_ai_v2/asr.py
```diff
@@ -727,21 +742,32 @@ def run_asr(
...
+    filename = "yt_audio"
+    content_type = "audio/wav"
+
     if is_yt_url(audio_url):
-        audio_url, size = download_youtube_to_wav_url(audio_url)
+        audio_wav_data = download_youtube_to_wav(audio_url)
     elif is_gdrive_url(furl(audio_url)):
         meta: dict[str, str] = gdrive_metadata(url_to_gdrive_file_id(furl(audio_url)))
+        filename = meta.get("name", "gdrive_audio")
+        content_type = meta.get("mimeType", "audio/wav")
```
devxpy, line 754: this looks wrong, the cotnent type of the uploaded audio is always wav at the end

daras_ai_v2/asr.py
```diff
@@ -1027,11 +1075,7 @@ def audio_url_to_wav(audio_url: str) -> tuple[str, int]:
         raise_for_status(r, is_user_url=True)
 
     wavdata, size = audio_bytes_to_wav(r.content)
-    if not wavdata:
```
devxpy, line 1030: Need to handle this case

devxpy: given how complex this is getting, it might be easier to ship the inefficient version that uploaded and downloaded the file for ghana nlp

daras_ai_v2/asr.py
```diff
@@ -1014,7 +1055,14 @@ def download_youtube_to_wav(youtube_url: str) -> bytes:
     return wavdata
 
 
-def audio_url_to_wav(audio_url: str) -> tuple[str, int]:
+def audio_url_to_wav_url(bytes_data: bytes, audio_url: str) -> tuple[str, int]:
```
devxpy, line 1058: this function doesnt do what it says, it should either be called audio_bytes_to_wav_url or actually do the thing it says it does

devxpy:
closing for now as the rearchitect around wav upload/download is getting complex. please raise the other inefficient version!

PR #461, opened 2024-09-10
Add GHANA NLP TTS API v1

anish-work:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

---
  - https://app.asana.com/0/0/1205803182950231

### DIFF
--- a/daras_ai_v2/asr.py
+++ b/daras_ai_v2/asr.py
@@ -202,7 +202,7 @@ MMS_SUPPORTED = {
 GHANA_NLP_SUPPORTED = {'en': 'English', 'tw': 'Twi', 'gaa': 'Ga', 'ee': 'Ewe', 'fat': 'Fante', 'dag': 'Dagbani',
                        'gur': 'Gurene', 'yo': 'Yoruba', 'ki': 'Kikuyu', 'luo': 'Luo', 'mer': 'Kimeru'}  # fmt: skip
 GHANA_NLP_MAXLEN = 500
-
+GHANA_API_AUTH_HEADERS = {"Ocp-Apim-Subscription-Key": str(settings.GHANA_NLP_SUBKEY)}
 
 class AsrModels(Enum):
     whisper_large_v2 = "Whisper Large v2 (openai)"
--- a/recipes/TextToSpeech.py
+++ b/recipes/TextToSpeech.py
@@ -409,7 +409,7 @@ class TextToSpeechPage(BasePage):
                     headers=GHANA_API_AUTH_HEADERS,
                     json={
                         "text": text,
-                        "language": gui.session_state.get(
+                        "language": state.get(
                             "ghana_nlp_tts_language",
                             GHANA_NLP_TTS_LANGUAGES.tw.name,
                         ),
--- a/recipes/TextToSpeech.py
+++ b/recipes/TextToSpeech.py
@@ -27,6 +27,7 @@ from daras_ai_v2.text_to_speech_settings_widgets import (
     elevenlabs_init_state,
 )
 import uuid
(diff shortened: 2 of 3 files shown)


recipes/TextToSpeech.py
```diff
@@ -401,6 +402,22 @@ def run(self, state: dict):
                 state["audio_url"] = upload_file_from_bytes(
                     "openai_tts.mp3", response.content
                 )
+            case TextToSpeechProviders.GHANA_NLP:
+                data = requests.post(
```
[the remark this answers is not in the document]
anish-work: @devxpy updated

recipes/TextToSpeech.py
```diff
@@ -401,6 +403,18 @@ def run(self, state: dict):
                 state["audio_url"] = upload_file_from_bytes(
                     "openai_tts.mp3", response.content
                 )
+            case TextToSpeechProviders.GHANA_NLP:
+                response = requests.post(
+                    "https://translation-api.ghananlp.org/tts/v1/tts",
+                    headers=GHANA_API_AUTH_HEADERS,
+                    json={
+                        "text": text,
+                        "language": "tw",
```
devxpy, line 412: Shouldnt this use the language that the user selected

daras_ai_v2/asr.py
```diff
@@ -202,7 +202,7 @@
 GHANA_NLP_SUPPORTED = {'en': 'English', 'tw': 'Twi', 'gaa': 'Ga', 'ee': 'Ewe', 'fat': 'Fante', 'dag': 'Dagbani',
                        'gur': 'Gurene', 'yo': 'Yoruba', 'ki': 'Kikuyu', 'luo': 'Luo', 'mer': 'Kimeru'}  # fmt: skip
 GHANA_NLP_MAXLEN = 500
-
+GHANA_API_AUTH_HEADERS = {"Ocp-Apim-Subscription-Key": str(settings.GHANA_NLP_SUBKEY)}
```
devxpy, line 205: This same constant should also be used in asr.py

recipes/TextToSpeech.py
```diff
@@ -57,7 +59,7 @@ class TextToSpeechSettings(BaseModel):
 
     openai_voice_name: OpenAI_TTS_Voices.api_choices | None
     openai_tts_model: OpenAI_TTS_Models.api_choices | None
-
+    ghana_tts_language: GHANA_TTS_LANGUAGES.api_choices | None
```
devxpy, line 62: Please change to `ghana_nlp_tts_language` given "ghana tts" doesnt show the right thing when you google vs "ghana nlp tts" and also for consistency with the rest of the codebase that uses `ghana_nlp_*` prefix

daras_ai_v2/asr.py
anish-work: @devxpy I added that in the other PRs too so I thought let's not do the same here

recipes/TextToSpeech.py
anish-work: @devxpy updated

devxpy: This should read from local var state not global gui.session state

recipes/TextToSpeech.py
anish-work: @devxpy updated

recipes/TextToSpeech.py
anish-work: @devxpy updated used the state var but I noticed that open is using gui.session_state so I used the same

devxpy:
lgtm, but please make sure you test localhost:3000/lipsynctts and localhost:3000/copilot on ghana nlp tts

PR #468, opened 2024-09-21
Add workspace functionality

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


routers/account.py
```diff
@@ -273,21 +350,25 @@ def _render_run(pr: PublishedRun):
...
 @contextmanager
-def account_page_wrapper(request: Request, current_tab: TabData):
+def account_page_wrapper(request: Request, current_tab: AccountTabs):
     if not request.user or request.user.is_anonymous:
         next_url = request.query_params.get("next", "/account/")
         redirect_url = furl("/login", query_params={"next": next_url})
         raise gui.RedirectException(str(redirect_url))
 
-    with page_wrapper(request):
+    with page_wrapper(request, route_fn=current_tab.route):
```
devxpy, line 364: pass down `current_tab` and build the redirect url from the tab. Passing the route_fn is very confusing

nikochiko: thank you, I particularly wanted review here.

my reasoning here to choose route was that tabs `AccountTabs` are scoped down to only the account page, whereas we use routes everywhere.
`page_wrapper` is used for all routes on the app, and we could scale this to other routes besides `AccountTabs`.

nikochiko:
@devxpy PR is still draft, I haven't added the workspace dropdown in publish dialog. API keys are being created with workspace_id but still charged with the uid.

I added here for review on whatever is implemented (or we could deploy it for admins only)

workspaces/widgets.py
```diff
@@ -60,3 +80,47 @@ def get_current_workspace(user: AppUser, session: dict) -> "Workspace":
...
+        profile_route,
+        workspaces_route,
+        workspaces_members_route,
+    )
+
+    if workspace.is_personal and route_fn in (
+        workspaces_members_route,
+        workspaces_route,
+    ):
+        # personal workspaces don't have a members page
```
devxpy, line 102: This redirect needs to be inside these routes themselves, not here

nikochiko:
I don't have access to this firestore collection https://console.firebase.google.com/u/0/v1/r/project/gooey-test-runner-1/firestore/indexes?create_composite=ClRwcm9qZWN0cy9nb29leS10ZXN0LXJ1bm5lci0xL2RhdGFiYXNlcy8oZGVmYXVsdCkvY29sbGVjdGlvbkdyb3Vwcy9hcGlfa2V5cy9pbmRleGVzL18QARoQCgx3b3Jrc3BhY2VfaWQQARoOCgpjcmVhdGVkX2F0EAEaDAoIX19uYW1lX18QAQ&pli=1

would need an index for `workspace_id` for api keys (i added this on the main firestore)

routers/account.py
devxpy: Lets not do that premature optimization and keep things simple. You're using the `route_fn` to do redirects based on some state, and those should really be inside the routes themselves. The usage for route_fn everywhere else is only to reverse URLs, something django for eg does via a special name param on every route (and we use the route's fn name as a proxy for that)

workspaces/widgets.py
nikochiko: The problem is that we end up with two sources of truth - the `session_state` and the URL path params.

e.g. you're on `/workspaces/devs-team-workspace-0A/saved/` and change the workspace from this selector to `Personal Account`.

When rendering workspace `saved` tab, we will see that the URL belongs to the workspace. We also end up at this same state when I share a `/workspaces/devs-team-workspace-0A/saved/` link with you were on `Personal Account` according to `gui.session_state`.

routers/account.py
nikochiko: That makes sense

workspaces/widgets.py
nikochiko: The intended behaviour is first case is that your `Personal Account` is selected and redirected to `/account/saved`, and in the second is that the URL overrides `gui.session_state`.

I created this other redirect path so that we can deal with the first case differently.

nikochiko: So the URL path params will always override `gui.session_state` if they have a conflict.

The selector widget can redirect / link to the other workspace's URL if a change needs to be made.

bots/migrations/0084_botintegration_workspace.py
```diff
@@ -0,0 +1,20 @@
...
+        ('workspaces', '0002_alter_workspace_domain_name'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='botintegration',
+            name='workspace',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, related_name='botintegrations', to='workspaces.workspace'),
+        ),
+    ]
```
nikochiko, line 20: to be superseded by https://github.com/GooeyAI/gooey-server/pull/500/files

bots/models.py
```diff
@@ -508,6 +541,12 @@ class BotIntegration(models.Model):
         help_text="The gooey account uid where the credits will be deducted from",
         db_index=True,
     )
+    workspace = models.ForeignKey(
+        "workspaces.Workspace",
+        on_delete=models.CASCADE,
+        related_name="botintegrations",
+        null=True,
+    )
```
nikochiko, line 549: to be superseded by https://github.com/GooeyAI/gooey-server/pull/500/files

nikochiko:
tests are finally green 🟢

daras_ai_v2/base.py
```diff
@@ -382,12 +383,28 @@ def _render_header(self):
...
+                                '<i class="fa-light fa-chevron-right"></i>',
+                                className="text-muted d-block mx-2",
+                                unsafe_allow_html=True,
+                            )
+                            gui.write(
+                                self.current_sr_user.full_name(),
+                                className="d-block text-muted me-2",
+                            )
+                    else:
+                        self.render_author(self.current_sr_user)
```
nikochiko, line 407: breadcrumbs - split this off

workspaces/widgets.py
```diff
@@ -182,3 +183,132 @@ def get_current_workspace(user: AppUser, session: dict) -> "Workspace":
...
+                options=[email_domain],
+                format_func=lambda s: s and f"@{s}" or "&mdash;" * 3,
+                value=email_domain,
+                allow_none=True,
+            )
+
+    if ref.pressed_confirm:
+        if workspace.domain_name:
+            workspace.save(update_fields=["domain_name"])
+        ref.set_open(False)
```
nikochiko, line 314: skip for now

nikochiko: look at how django does this

workspaces/models.py
```diff
@@ -127,6 +143,8 @@ class Workspace(SafeDeleteModel):
 
     objects = WorkspaceQuerySet.as_manager()
 
+    api_hashids = hashids.Hashids(salt=settings.HASHIDS_API_SALT + "/workspaces")
```
nikochiko, line 146: get rid of this

PR #491, opened 2024-10-17
Migrate API keys from firebase

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


api_keys/admin.py
```diff
@@ -0,0 +1,8 @@
+from django.contrib import admin
+
+from .models import ApiKey
+
+
+@admin.register(ApiKey)
+class ApiKeyAdmin(admin.ModelAdmin):
+    list_display = ("preview", "workspace", "created_by", "created_at")
```
nikochiko, line 8: `search_fields` - workspace__name, workspace__id, created_by__display_name, created_by__uid, preview

nikochiko: + do some research - how we do it in other places

api_keys/models.py
```diff
@@ -0,0 +1,67 @@
...
+                preview=safe_preview(secret_key),
+                workspace=workspace,
+                **kwargs,
+            ),
+        )[0]
+
+    def get_from_secret_key(self, secret_key: str) -> "ApiKey | None":
+        hasher = PBKDF2PasswordHasher()
+        hash = hasher.encode(secret_key)
+        return self.filter(hash=hash).first()
```
nikochiko, line 45: `self.get(hash=hash)` instead

api_keys/models.py
```diff
@@ -0,0 +1,67 @@
...
+        return self.get_or_create(
+            hash=hasher.encode(secret_key),
+            defaults=dict(
+                preview=safe_preview(secret_key),
+                workspace=workspace,
+                **kwargs,
+            ),
+        )[0]
+
+    def get_from_secret_key(self, secret_key: str) -> "ApiKey | None":
```
nikochiko, line 42: `typing.Optional["ApiKey"]`

scripts/migrate_workspaces.py
```diff
@@ -100,3 +149,23 @@ def update_in_batches(qs, **kwargs):
...
+
+
+@contextmanager
+def disable_auto_now_add(model, field_name):
+    for field in model._meta.local_fields:
+        if field.name == field_name:
+            field.auto_now_add = False
+            yield
+            field.auto_now_add = True
+            break
```
nikochiko, line 171: get rid of this

scripts/migrate_workspaces.py
```diff
@@ -90,6 +96,49 @@ def migrate_published_runs():
...
+                        preview=snap.get("secret_key_preview"),
+                        workspace_id=cached_workspaces_by_uid[snap.get("uid")].id,
+                        created_by_id=cached_workspaces_by_uid[
+                            snap.get("uid")
+                        ].created_by.id,
+                        created_at=snap.get("created_at"),
+                        updated_at=snap.get("created_at"),
+                    )
+                    for snap in batch
+                    if snap.get("uid") in cached_workspaces_by_uid
```
nikochiko, line 133: `if workspace := cached_workspaces_by_uid.get(snap.get("uid"))`

nikochiko:
Add "Created By" to the table

nikochiko:
This PR does not have changes to authenticate workspaces instead of app users.

That would be a larger change and would need changes to parts that use the authentication code too, because of a different type signature (e.g. bot integrations).

nikochiko:
@devxpy these are the changes to make use of `Workspace` with API keys - in `submit_api_call` and then everywhere that is used (including Bot Integrations) - https://github.com/GooeyAI/gooey-server/pull/500/files

What do you think about rolling a first release to prod where we keep authenticating for `AppUser` in `auth/token_authentication.py` and migrate the keys to local?

Then we can figure out the usage of workspaces for everywhere that API keys are used.

PR #498, opened 2024-10-24
UI: add rect logo for mobile

anish-work:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/settings.py
+++ b/daras_ai_v2/settings.py
@@ -226,6 +226,7 @@ if not firebase_admin._apps:
     firebase_admin.initialize_app()
 
 GOOEY_LOGO_IMG = "https://storage.googleapis.com/dara-c1b52.appspot.com/daras_ai/media/2a3aacb4-0941-11ee-b236-02420a0001fb/thumbs/logo%20black.png_400x400.png"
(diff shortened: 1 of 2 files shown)


daras_ai_v2/settings.py
```diff
@@ -226,6 +226,7 @@
     firebase_admin.initialize_app()
 
 GOOEY_LOGO_IMG = "https://storage.googleapis.com/dara-c1b52.appspot.com/daras_ai/media/2a3aacb4-0941-11ee-b236-02420a0001fb/thumbs/logo%20black.png_400x400.png"
+GOOEY_LOGO_RECT = "https://storage.googleapis.com/dara-c1b52.appspot.com/daras_ai/media/d628be8a-9207-11ef-8aee-02420a000186/984x272%20rect%20gooey%20logo.png"
```
anish-work, line 229: @devxpy should we put the logo in some other folder than this ? I don't have access to the bucket please upload in the right folder if needed and give me the url I will change

devxpy: it gets resized automatically -

https://storage.googleapis.com/dara-c1b52.appspot.com/daras_ai/media/d628be8a-9207-11ef-8aee-02420a000186/thumbs/984x272%20rect%20gooey%20logo_400x400.png

devxpy:
lgtm

PR #499, opened 2024-10-28
Generate integration details

clr-li:
Asana: https://app.asana.com/0/1203067047205953/1208307422943341

### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


daras_ai_v2/bot_integration_widgets.py
```diff
@@ -391,14 +393,65 @@ def web_widget_config(bi: BotIntegration, user: AppUser | None):
...
-        bi.descripton = gui.text_area(
-            "###### Description",
-            value=bi.descripton,
+            gencol1, gencol2 = gui.columns(2)
+            with gencol1:
+                if gui.button(f"{icons.camera} Change Photo"):
+                    gui.session_state["--update-display-picture"] = True
+                    gui.rerun()
+            with gencol2:
+                generate_prompt = """Your job is to define the schema for a chatbot, using the title, description and prompt of an AI Workflow.
```
devxpy, line 402: isn't this prompt supposed to be in a /llm example?

clr-li: You're right, I removed the generate_prompt var

daras_ai_v2/bot_integration_widgets.py
```diff
@@ -391,14 +393,53 @@ def web_widget_config(bi: BotIntegration, user: AppUser | None):
...
+                                variables={"title": bi.name, "prompt": copilot_prompt},
+                            ),
+                        )
+                    )
+                    sr.wait_for_celery_result(result)
+                    output_dict = json.loads(sr.state["output_text"]["gpt_4_o"][0])
+                    gui.session_state["bi-description"] = output_dict["description"]
+                    for i in range(4):
+                        gui.session_state[f"--question-{i}"] = output_dict[
+                            "User_question_" + str(i)
```
devxpy, line 433: You want to make this schema a list in the json returned. This is a bad way to encode information that you want to index later as an array

PR #500, opened 2024-10-29
Add workspaces to API auth and bot integrations

nikochiko:
- token authentication should return Workspace
- Use workspace for API calls
- sr.submit_api_call and pr.submit_api_call to use workspaces
- Add workspace to bot integrations
- Add migration script for bot integrations
- make use of workspace in integrations & functions

### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


bots/tasks.py
```diff
@@ -89,7 +84,8 @@ def msg_analysis(self, msg_id: int, anal_id: int, countdown: int | None):
 
     # make the api call
     result, sr = analysis_sr.submit_api_call(
-        current_user=billing_account,
+        workspace=msg.conversation.bot_integration.workspace,
+        current_user=msg.conversation.bot_integration.created_by,
```
nikochiko, line 88: i think the `current_user` here doesn't serve any purpose, and it might even be counterproductive - e.g. it would be difficult to separate bot-integration requests from rest of the API calls made by a particular user - useful for usage/limits calculation.

also we always have access to `bot_integration.created_by` through a join.

devxpy: see other usages: https://github.com/GooeyAI/gooey-server/blob/bf3e66e06eceacbbdfc94cb8ad7dfd0a8ec76736/routers/api.py#L158-L172

PR #502, opened 2024-11-06
Add Citation fragments

milovate:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/search_ref.py
+++ b/daras_ai_v2/search_ref.py
@@ -304,67 +304,10 @@ def render_output_with_refs(state, height=500):
         html = render_text_with_refs(text, state.get("references", []))
         scrollable_html(html, height=height)
 
-
 FOOTNOTE_SYMBOLS = ["*", "†", "‡", "§", "¶", "#", "♠", "♥", "♦", "♣", "✠", "☮", "☯", "✡"]  # fmt: skip
 def generate_footnote_symbol(idx: int) -> str:
     quotient, remainder = divmod(idx, len(FOOTNOTE_SYMBOLS))
     return FOOTNOTE_SYMBOLS[remainder] * (quotient + 1)
-
-
-# def generate_text_fragment(text, min_len=0, max_len=30):
-#     """
-#     Generate a text fragment for a given text.
-
-#     Args:
-#         text (str): The input text to generate a fragment from.
-#         min_len (int): The minimum length for the prefix and suffix.
-#         max_len (int): The maximum length for the prefix and suffix.
-
-#     Returns:
-#         str: The encoded text fragment.
-#     """
-
-#     def find_alphanumeric_window(s, window_size, reverse=False):
-#         range_func = (
-#             range(len(s) - window_size + 1, 0, -1)
-#             if reverse
-#             else range(len(s) - window_size + 1)
-#         )
-#         for i in range_func:
-#             window = s[i : i + window_size]
-#             # Skip windows containing newline characters
-#             if "\n" in window:
-#                 continue
-#             print(window)
-#             # Updated regex to include allowed symbols and exclude '[', ']', and '.'
-#             if re.match(r'^[a-zA-Z0-9\s"\'.,?:;()&^*$@!#]+$', window) and not re.search(
-#                 r"[\[\].]", window
-#             ):
-#                 words = window.strip().split()
-#                 if (
-#                     words
-#                     and re.match(r"^[a-zA-Z0-9]+$", words[0])
-#                     and re.match(r"^[a-zA-Z0-9]+$", words[-1])
-#                 ):
-#                     return window.strip()
-#         # Fallback if no suitable window is found
-#         return re.sub(r"[\[\].]", "", s[:window_size]).strip()
-
-#     prefix = find_alphanumeric_window(text, max_len)
-#     suffix = find_alphanumeric_window(text, max_len, reverse=True)
-
-#     # Ensure prefix and suffix meet minimum length requirements
-#     if len(prefix) < min_len:
-#         prefix = find_alphanumeric_window(text, min_len)
-#     if len(suffix) < min_len:
-#         suffix = find_alphanumeric_window(text, min_len, reverse=True)
-#     prefix = " ".join(prefix.split()[:-1])
-#     suffix = " ".join(suffix.split()[1:])
-#     # Assemble and encode the fragment
-#     text_fragment = f"#:~:text={quote(prefix, safe='')},{quote(suffix, safe='')}"
-
-#     return text_fragment
-
 
 def find_alphanumeric_window(s, window_size, reverse=False):
     range_func = (
--- a/daras_ai_v2/search_ref.py
+++ b/daras_ai_v2/search_ref.py
@@ -8,7 +8,8 @@ from typing_extensions import TypedDict
 from daras_ai_v2.exceptions import UserError
 from daras_ai_v2.scrollable_html_widget import scrollable_html
 
-
+from loguru import logger
+from furl import furl
 from urllib.parse import quote
 import re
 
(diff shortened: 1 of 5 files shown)


daras_ai_v2/search_ref.py
```diff
@@ -302,3 +305,47 @@ def render_output_with_refs(state, height=500):
...
+                continue
+            # Updated regex to include allowed symbols and exclude '[', ']', and '.'
+            if re.match(r'^[a-zA-Z0-9\s"\'.,?:;()&^*$@!#]+$', window) and not re.search(r'[\[\].]', window):
+                words = window.strip().split()
+                if words and re.match(r'^[a-zA-Z0-9]+$', words[0]) and re.match(r'^[a-zA-Z0-9]+$', words[-1]):
+                    return window.strip()
+        # Fallback if no suitable window is found
+        return re.sub(r'[\[\].]', '', s[:window_size]).strip()
+
+    prefix = find_alphanumeric_window(text, max_len)
```
devxpy, line 338: I don't think you're passing the max_len here correctly

daras_ai_v2/search_ref.py
```diff
@@ -302,3 +305,47 @@ def render_output_with_refs(state, height=500):
...
+
+    Args:
+        text (str): The input text to generate a fragment from.
+        min_len (int): The minimum length for the prefix and suffix.
+        max_len (int): The maximum length for the prefix and suffix.
+
+    Returns:
+        str: The encoded text fragment.
+    """
+    def find_alphanumeric_window(s, window_size, reverse=False):
```
devxpy, line 323: This doesn't seem like it needs to be a locally scoped function, use top level scope

daras_ai_v2/search_ref.py
```diff
@@ -302,3 +305,47 @@ def render_output_with_refs(state, height=500):
...
+
+    # Ensure prefix and suffix meet minimum length requirements
+    if len(prefix) < min_len:
+        prefix = find_alphanumeric_window(text, min_len)
+    if len(suffix) < min_len:
+        suffix = find_alphanumeric_window(text, min_len, reverse=True)
+    prefix = ' '.join(prefix.split()[:-1])
+    suffix = ' '.join(suffix.split()[1:])
+    # Assemble and encode the fragment
+    text_fragment = f"#:~:text={quote(prefix, safe='')},{quote(suffix, safe='')}"
```
devxpy, line 349: please use furl for url consturction as discussed earlier

milovate: furl automatically encodes the `:~:` which should remain unencoded , couldn't find any work around except building url manually

daras_ai_v2/search_ref.py
milovate: yes, will do

daras_ai_v2/search_ref.py
devxpy: Oops

devxpy: Lets do this then, just in case the URL has a fragment already - 

```
return str(furl(url).remove(fragment=True)) + "#:~:text=..."
```

daras_ai_v2/search_ref.py
```diff
@@ -297,8 +304,62 @@ def render_output_with_refs(state, height=500):
         html = render_text_with_refs(text, state.get("references", []))
         scrollable_html(html, height=height)
 
-
 FOOTNOTE_SYMBOLS = ["*", "†", "‡", "§", "¶", "#", "♠", "♥", "♦", "♣", "✠", "☮", "☯", "✡"]  # fmt: skip
 def generate_footnote_symbol(idx: int) -> str:
     quotient, remainder = divmod(idx, len(FOOTNOTE_SYMBOLS))
     return FOOTNOTE_SYMBOLS[remainder] * (quotient + 1)
+
+def find_alphanumeric_window(s, window_size, reverse=False):
```
devxpy, line 312: I'll definitely say Bravo! for figuring this out. It seems to work pretty well. 

But at the same time it is a bit complex and prone to failures + is compute inefficient because of the n x m string window loops.

https://developer.mozilla.org/en-US/docs/Web/URI/Fragment/Text_fragments

I did some digging and realised you don't need both start and end strings, and can only do the start string. on top of that, you can provide multiple of these strings. It feels like we can build redundancy using this method, and can do it entirely in regex and avoid slower python loops given the requirement as per sean is to only be able to scroll to the relevant section, not necessarily highlight everything.


E.g. for this chunk:

```
TV Movie

Werewolf by Night (art)

Web Series

WHiH Newsfront (footage)

Docuseries

Legends
*The Avengers Initiative

Game

Avengers: Damage Control

Comic

Iron Man 3 Prelude
Thor: The Dark World Prelude
Captain America: Homecoming (mentioned)
The Avengers Adaptation
Avengers: Age of Ultron Prelude - This Scepter'd Isle
Avengers: Operation HYDRA
Captain America: Civil War Prelude Infinite Comic
Captain America: Road to War
Spider-Man: Homecoming Prelude
Avengers: Infinity War Prelude
Ant-Man and the Wasp Prelude
Captain Marvel Prelude
Avengers: Endgame Prelude
Spider-Man: Far From Home Prelude
Black Widow Prelude (flashbacks)

Book

Avengers: Infinity War: The Heroes' Journey
Captain Marvel: What Makes A Hero
The Wakanda Files: A Technological Exploration of the Avengers and
Beyond
Look Out for the Little Guy! (mentioned)

"The Avengers. That's what we call ourselves. We're sort of like a team.
'Earth's Mightiest Heroes'-type thing."

―Tony Stark to Loki^([src])

The Avengers were a team of extraordinary individuals, with either
superpowers or other special characteristics. Though primarily
affiliated with the interests of the United States of America, the
group's purpose was to protect global stability from inner or
extraterrestrial threats. The Avengers were first assembled by
S.H.I.E.L.D. as a result of the Avengers Initiative, when Loki invaded
Earth with his Chitauri army. The team, consisting of Iron Man, Captain
America, Hulk, Thor, Black Widow and Hawkeye defeated Loki and went
their separate ways for a while.
```

current code gives: https://marvelcinematicuniverse.fandom.com/wiki/Avengers#:~:text=Captain%20America%3A%20Civil%20War,separate%20ways%20for%20a%20while


But if we do the other method of using multiple sentences (joined together using `&text=`) then it sorta works:

https://marvelcinematicuniverse.fandom.com/wiki/Avengers#:~:text=Avengers:%20Age%20of%20Ultron%20Prelude%20-%20This%20Scepter'd%20Isle&text=Captain%20Marvel%20Prelude&text=The%20Wakanda%20Files:%20A%20Technological%20Exploration%20of%20the%20Avengers%20and&text=The%20Avengers%20were%20a%20team%20of%20extraordinary%20individuals&text=Look%20Out%20for%20the%20Little%20Guy!&text=The%20Avengers%20were%20a%20team%20of%20extraordinary%20individuals

daras_ai_v2/search_ref.py
milovate: It would break the pdf fragment for #page= 
we can do this selectively -
 
```
    url_obj = furl(url)
    if not url_obj.path.segments[-1].endswith('.pdf'):
        url_obj.remove(fragment=True)
    
    return str(url_obj) + text_fragment
```

devxpy: I believe for PDF we shouldn't do text fragment at all given it doesn't work. Plus, instead of checking for .pdf, we should do the inverse and check for html content type in `EmbeddingsReference.embedded_file.metadata.mime_type` here so we don't give the text fragment in non html pages (like slides docx etc)

https://github.com/GooeyAI/gooey-server/blob/51407ec9e94c4838e1036d010209e880f9e368a1/daras_ai_v2/vector_search.py#L224

PR #509, opened 2024-11-12
workspace selectbox in publish form

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


app_users/models.py
```diff
@@ -233,6 +234,16 @@ def get_or_create_personal_workspace(self) -> tuple["Workspace", bool]:
...
 
+    @cached_property
+    def cached_workspaces(self) -> list["Workspace"]:
+        from workspaces.models import Workspace
+
+        return list(
+            Workspace.objects.filter(
+                memberships__user=self, memberships__deleted__isnull=True
+            ).order_by("-is_personal", "-created_at")
+        ) or [self.get_or_create_personal_workspace()[0]]
```
nikochiko, line 245: more efficient would be to make use of `self.memberships`, and use `AppUser.objects.prefetch_related("memberships__workspace")` when retrieving the user.

we also fetch the memberships to figure out the role in some cases. this can be `[m.workspace for m in self.memberships]`. this would be only 2 DB queries - one for appuser, and one for membership-workspace - throughout the request-response cycle. other than the runs.

ref: https://docs.djangoproject.com/en/5.1/ref/models/querysets/#prefetch-related

app_users/models.py
```diff
@@ -233,6 +234,16 @@ def get_or_create_personal_workspace(self) -> tuple["Workspace", bool]:
 
         return Workspace.objects.get_or_create_from_user(self)
 
+    @cached_property
```
nikochiko, line 237: django has its own `cached_property` decorator for use with models https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.functional.cached_property

nikochiko: difference here: https://stackoverflow.com/a/68594071

we want thread safety and work in python <3.12, django's is more efficient there

app_users/models.py
devxpy: makes sense, lets do that too

daras_ai_v2/base.py
```diff
@@ -569,7 +569,22 @@ def _render_share_modal(self, dialog: gui.AlertDialogRef):
...
+                duplicate = gui.button(
+                    f"{icons.fork} Duplicate", type="link", className="d-inline m-0 p-0"
+                )
+                gui.html(" this workflow to edit with others")
+                ref = gui.use_alert_dialog(key="publish-modal")
+                if duplicate:
+                    self.clear_publish_form()
+                    ref.set_open(True)
+                if ref.is_open:
+                    gui.session_state["published_run_workspace"] = workspaces[-1].id
```
nikochiko, line 584: @devxpy having this assignment here creates the bug that no other workspace can be selected, because we overwrite it and `ref.is_open` is run every time after `ref.set_open(True)`.

we only want this to be the default - hence I had put this in the `if duplicate: ...` branch

PR #515, opened 2024-11-14
fix header on workflow pages: breadcrumb alignment, title, and responsive share buton

nikochiko:
- **fix alignment in header on workflow pages**
- **add text-nowrap on "unpublished changes" indicator**
- **fix bug where 'Share' modal disappears when form input changes**
- **hide text on 'share' button on smaller screens**
- **fix: make title occupy full-width under breadcrumbs+buttons row**

### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/base.py
+++ b/daras_ai_v2/base.py
@@ -414,7 +414,7 @@ class BasePage:
         if not workspace:
             return
 
-        with gui.breadcrumbs(divider="▶"):
+        with gui.breadcrumbs(divider=""):
             with gui.tag("li", className="breadcrumb-item"):
                 self.render_workspace_author(workspace)
 
@@ -426,6 +426,9 @@ class BasePage:
                 "li",
                 className="breadcrumb-item d-flex align-items-center container-margin-reset",
             ):
+                with gui.tag("small", className="text-muted pe-2"):
+                    gui.html(icons.chevron_right)
+
                 if user := self.current_sr_user:
                     full_name = user.full_name()
                     link = user.handle and user.handle.get_app_url()
--- a/daras_ai_v2/base.py
+++ b/daras_ai_v2/base.py
@@ -368,7 +368,9 @@ class BasePage:
         request_changed = self._has_request_changed()
 
         with gui.div(className="d-flex justify-content-between align-items-start mt-4"):
-            with gui.div():
+            left, right = gui.div(), gui.div()
+
+            with left:
                 if tbreadcrumbs.has_breadcrumbs():
                     with gui.div(
                         className="d-block d-lg-flex align-items-center pt-2 mb-2"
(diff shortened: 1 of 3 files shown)


daras_ai_v2/base.py
```diff
@@ -367,30 +367,34 @@ def _render_header(self):
         can_save = self.can_user_save_run(sr, pr)
         request_changed = self._has_request_changed()
 
-        with gui.div(className="d-flex justify-content-between mt-4"):
-            with gui.div(className="d-lg-flex d-block align-items-center"):
-                if not tbreadcrumbs.has_breadcrumbs() and not self.current_sr_user:
-                    self._render_title(tbreadcrumbs.h1_title)
-
-                if tbreadcrumbs:
```
nikochiko, line 375: the extra heading was a regression. this if condition was busted - `tbreadcrumbs.has_breadcrumbs` needs to be used.

daras_ai_v2/base.py
```diff
@@ -367,30 +367,34 @@ def _render_header(self):
         can_save = self.can_user_save_run(sr, pr)
         request_changed = self._has_request_changed()
 
-        with gui.div(className="d-flex justify-content-between mt-4"):
-            with gui.div(className="d-lg-flex d-block align-items-center"):
-                if not tbreadcrumbs.has_breadcrumbs() and not self.current_sr_user:
```
nikochiko, line 372: removed `not self.current_sr_user` check because breadcrumbs and user always exist together.

cases:
- root workflow - no author, no breadcrumbs
- example - both author and breadcrumbs
- saved run - both author and breadcrumbs

daras_ai_v2/base.py
```diff
@@ -410,7 +414,7 @@ def _render_author_as_breadcrumb(self, *, is_example: bool, is_root_example: boo
         if not workspace:
             return
 
-        with gui.breadcrumbs(divider="▶"):
+        with gui.breadcrumbs(divider=""):
```
nikochiko, line 417: i tried adding the SVG here as the divider - https://getbootstrap.com/docs/5.2/components/breadcrumb/

but it is tricky to get right - SVG has to be encoded correctly and we would need to change the code in `gooey-gui` too to make that work - to remove single quotes we put before adding it to `style=...`. this is a simple workaround for a rare case. we keep icons simple & consistent in html, and avoid complexity of conforming to the data-URI and encoding that is needed.

daras_ai_v2/base.py
```diff
@@ -491,7 +497,7 @@ def render_social_buttons(self):
                     padding: 6px;
                 }
                 </style>
-                """
+                """.strip()
```
nikochiko, line 500: to remove extra white space from the text

devxpy:
lgtm rebase and merge pls

PR #519, opened 2024-11-15
add pptx support ,google slide fragment, azure form reco for pptx  

milovate:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


daras_ai_v2/vector_search.py
```diff
@@ -823,7 +834,99 @@ def tabular_bytes_to_any_df(
...
+
+    if shape.shape_type == MSO_SHAPE_TYPE.GROUP:
+        for sub_shape in shape.shapes:
+            check_recursively_for_text(sub_shape, group_text)
+    else:
+        if hasattr(shape, "text"):
+            text = shape.text.strip()
+            if text:
+                group_text.append(f"{text}")
+
```
milovate, line 928: Referenced this StackOverflow post for implementing a recursive code block:  
https://stackoverflow.com/questions/51701626/how-to-extract-text-from-a-text-shape-within-a-group-shape-in-powerpoint-using

daras_ai_v2/azure_doc_extract.py
```diff
@@ -77,6 +89,18 @@ def azure_form_recognizer(url: str, model_id: str, params: dict = None):
             case _:
                 sleep(1)
 
+def extract_records_prebuilt_read(result: dict, page_num: int) -> list[dict]:
```
devxpy, line 92: instead of creating a new func, update extract_records to handle both cases - eg I think its because the result won't have tables, so do `result.get('tables', [])` instead of `result['tables']`

daras_ai_v2/azure_doc_extract.py
```diff
@@ -31,11 +34,20 @@ def azure_doc_extract_page_num(pdf_url: str, page_num: int) -> str:
...
+            records_to_text(extract_records_prebuilt_read(result, page["pageNumber"]))
+            for page in result["pages"]
+        ]
+    else:
+        
+        result = azure_form_recognizer(pdf_url, model_id, params, azure_service="formrecognizer")
+        return [
+            records_to_text(extract_records(result, page["pageNumber"]))
+            for page in result["pages"]
+        ]
```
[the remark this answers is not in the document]
devxpy: that's ok but why do we use the new api for prebuilt-read only?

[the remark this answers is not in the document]
devxpy: and i would also not prefer to use the `-preview` model if not necessary

daras_ai_v2/gdrive_downloader.py
```diff
@@ -9,6 +9,8 @@
 def is_gdrive_url(f: furl) -> bool:
     return f.host in ["drive.google.com", "docs.google.com"]
 
+def is_gdrive_presentation_url(f: furl) -> bool:
+    return is_gdrive_url(f) and "presentation" in f.path.segments
```
devxpy, line 13: hmm, this looks wrong, you only want docs.google.com URLs here no?  drive.google.com implies a pptx uploaded to google drive which is different  from a docs.google.com google slides url

daras_ai_v2/vector_search.py
```diff
@@ -873,6 +1001,11 @@ def get_pdf_num_pages(f_bytes: bytes) -> int:
 
 
 def add_page_number_to_pdf(url: str | furl, page_num: int) -> furl:
+
+    # if it's a google drive presentation, add the slide number to the fragment
+    if is_gdrive_presentation_url(furl(url)):
+        return furl(url).set(fragment_args={"slide": page_num} if page_num else {})
+
     return furl(url).set(fragment_args={"page": page_num} if page_num else {})
```
devxpy, line 1009:
```
if is_gdrive_presentation_url(...):
	param = "slide"
else:
	param = "page"
return furl(..., {param: page_num ...})
```

daras_ai_v2/vector_search.py
```diff
@@ -823,7 +834,99 @@ def tabular_bytes_to_any_df(
...
+    """
+    Extracts and converts text, tables, charts, and grouped shapes from a PPTX file into Markdown format.
+    """
+    from pptx import Presentation
+    from pptx.enum.shapes import MSO_SHAPE_TYPE
+
+    prs = Presentation(f)
+    slides_text = []
+
+    for slide_idx, slide in enumerate(prs.slides, start=1):
```
devxpy, line 847: slide_num , not slide_idx 😅

daras_ai_v2/gdrive_downloader.py
milovate: Yes, only docs.google.com supports scrolling directly to the desired slide. I initially thought appending #slide=4 to the URL would provide meaningful context for users when sharing/reading a Google Drive preview link for a PowerPoint file.  example:
https://drive.google.com/file/d/1abcdef/view#slide=4 – but this doesn't scroll down to the slide 4

PR #536, opened 2024-11-27
bugfix: only invite new users to workspaces with matching domain, accept all pending invitations on AppUser->post_save

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

---
  - https://app.asana.com/0/0/1208851302477764

### DIFF
--- a/workspaces/signals.py
+++ b/workspaces/signals.py
@@ -1,33 +1,47 @@
-import traceback
-from django.db.models.signals import post_save
+from django.core.exceptions import ValidationError
+from django.db.models.signals import post_save, pre_save
 from django.dispatch import receiver
 from loguru import logger
 from safedelete.signals import post_softdelete
-import sentry_sdk
 
 from app_users.models import AppUser
+from workspaces.tasks import send_added_to_workspace_email
 from .models import Workspace, WorkspaceInvite, WorkspaceMembership, WorkspaceRole
 
 
-@receiver(post_save, sender=AppUser)
-def add_user_existing_workspace(instance: AppUser, **kwargs):
-    """
-    if the domain name matches, add the user to the workspace
-    """
-    if not instance.email:
+@receiver(pre_save, sender=AppUser)
+def invite_user_to_workspaces_with_matching_domain(instance: AppUser, **kwargs):
+    if instance.id or not instance.email:
+        # don't send invitation if: 1. user is being updated, 2. user doesn't have an email
         return
+
     email_domain = instance.email.split("@")[-1].lower()
     for workspace in Workspace.objects.filter(domain_name=email_domain):
+        WorkspaceInvite.objects.get_or_create(
+            workspace=workspace,
+            email=instance.email,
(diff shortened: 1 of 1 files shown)


devxpy:
I thought we discussed not doing the pre/post signal here - doesn't make a lot of sense to me

devxpy:
image

nikochiko:
Ah, sorry. I misread it as "does make a lot of sense". Will revert

workspaces/signals.py
```diff
@@ -1,33 +1,47 @@
...
-@receiver(post_save, sender=AppUser)
-def add_user_existing_workspace(instance: AppUser, **kwargs):
-    """
-    if the domain name matches, add the user to the workspace
-    """
-    if not instance.email:
+@receiver(pre_save, sender=AppUser)
+def invite_user_to_workspaces_with_matching_domain(instance: AppUser, **kwargs):
+    if instance.id or not instance.email:
+        # don't send invitation if: 1. user is being updated, 2. user doesn't have an email
```
devxpy, line 15: this does not allow existing users to be added automatically?

nikochiko: that we do in `post_save`

devxpy: the invite will not exist?

nikochiko: in any case - that code would not be in an `AppUser` signal, but after a workspace was updated

devxpy: e.g. I think this wont work.
1. create a workspace
2. signup as dev@dara.network
3. update workspace domain to dara.network
4. login as dev@dara.network 
5. ????

nikochiko:
@devxpy but then - this signal would run for users when they e.g. update their profile. that seems very random

workspaces/signals.py
nikochiko: that wouldn't have worked before this PR too

devxpy: Everytime you login, it saves the user so the old check would run no?

nikochiko: also fixes the case when:
1. user is added to a workspace because domain name
2. user leaves workspace
3. user updates profile -> user is added to the workspace again

nikochiko: and of course, doesn't use `create_and_send_invite` incorrectly

nikochiko:
superseded by https://github.com/GooeyAI/gooey-server/pull/546

workspaces/signals.py
nikochiko: > Everytime you login, it saves the user so the old check would run no?

why would login save the user?

devxpy: Ok - I thought we tracked user's last login but we apparently don't

devxpy:
Lets make sure both scenarios work - https://app.asana.com/0/1203067047205953/1208893638284128

PR #540, opened 2024-11-30
fix: pptx uploaded on google docs 

milovate:
### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/gdrive_downloader.py
+++ b/daras_ai_v2/gdrive_downloader.py
@@ -62,15 +62,25 @@ def gdrive_list_urls_of_files_in_folder(f: furl, max_depth: int = 4) -> list[str
 
 def gdrive_download(f: furl, mime_type: str) -> tuple[bytes, str]:
     from googleapiclient import discovery
-    from googleapiclient.http import MediaIoBaseDownload
-    from googleapiclient.errors import HttpError
 
     # get drive file id
     file_id = url_to_gdrive_file_id(f)
     # get metadata
     service = discovery.build("drive", "v3")
+
+    request, mime_type = service_request(service, file_id, f, mime_type)
+    file_bytes, mime_type = download_blob_file_content(
+        service, request, file_id, f, mime_type
+    )
+
+    return file_bytes, mime_type
+
+
+def service_request(
+    service, file_id: str, f: furl, mime_type: str, retried_request=False
+) -> tuple[any, str]:
     # get files in drive directly
-    if f.host == "drive.google.com":
+    if f.host == "drive.google.com" or retried_request:
         request = service.files().get_media(
             fileId=file_id,
             supportsAllDrives=True,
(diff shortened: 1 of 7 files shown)


devxpy:
I think we talked about exporting the file as presentation in the first place here
https://github.com/GooeyAI/gooey-server/blob/66a05d78bd85312747343e6b976b1c75f386af92/daras_ai_v2/gdrive_downloader.py#L159-L161

I would even go further to suggest that you likely don't need export_media() at all for presentation and just use get_media() always

devxpy:
> I think we talked about exporting the file as presentation in the first place here
> 
> https://github.com/GooeyAI/gooey-server/blob/66a05d78bd85312747343e6b976b1c75f386af92/daras_ai_v2/gdrive_downloader.py#L159-L161

Ok nevermind, I don't think that works. But we can be smart about it by looking at the mime type returned by `gdrive_metadata` (which we already have here)

devxpy:
In addition, we can also add `exportLinks` to the `gdrive_metadata` to get export slides larger than 10mb https://stackoverflow.com/a/59168288/7061265

daras_ai_v2/gdrive_downloader.py
```diff
@@ -60,16 +62,27 @@ def gdrive_list_urls_of_files_in_folder(f: furl, max_depth: int = 4) -> list[str
...
+    request, mime_type = service_request(service, file_id, f, mime_type)
+    file_bytes, mime_type = download_blob_file_content(
+        service, request, file_id, f, mime_type, export_links
+    )
+
+    return file_bytes, mime_type
+
+
+def service_request(
+    service, file_id: str, f: furl, mime_type: str, retried_request=False
```
devxpy, line 82: remove unused param

daras_ai_v2/gdrive_downloader.py
```diff
@@ -81,17 +94,60 @@ def gdrive_download(f: furl, mime_type: str) -> tuple[bytes, str]:
...
+            # print(f"Download {int(status.progress() * 100)}%")
+        f_bytes = file.getvalue()
+
     return f_bytes, mime_type
 
 
+def download_from_exportlinks(f: furl) -> bytes:
+    try:
+        r = requests.get(f)
+        f_bytes = r.content
```
devxpy, line 145: use raise_for_status() like we do everywhere else in code

devxpy: this likely doesnt support downloading private docs without auth, I may be wrong

daras_ai_v2/gdrive_downloader.py
```diff
@@ -81,17 +94,60 @@ def gdrive_download(f: furl, mime_type: str) -> tuple[bytes, str]:
...
+                # print(f"Download {int(status.progress() * 100)}%")
+            f_bytes = file.getvalue()
+
+    else:
+        done = False
+        while done is False:
+            _, done = downloader.next_chunk()
+            # print(f"Download {int(status.progress() * 100)}%")
+        f_bytes = file.getvalue()
+
```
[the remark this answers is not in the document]
milovate: ah! , yes this make way more sense, i wanted to only use export_links for pptx without breaking the existing logic hence the mess.

PR #547, opened 2024-12-03
gracefully handle missing savedrun for bot integration

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

---
  - https://app.asana.com/0/0/1208584814057545

### DIFF
--- a/daras_ai_v2/bots.py
+++ b/daras_ai_v2/bots.py
@@ -78,16 +78,16 @@ class BotInterface:
     user_id: str
     convo: Conversation
     bi: BotIntegration
-    saved_run: SavedRun
+    saved_run: SavedRun | None = None
     input_type: typing.Literal[
         "text", "audio", "video", "image", "document", "interactive"
     ]
-    user_msg_id: str = None
+    user_msg_id: str | None = None
     can_update_message: bool = False
 
-    page_cls: typing.Type[BasePage] = None
+    page_cls: typing.Type[BasePage] | None = None
     query_params: dict
-    user_language: str = None
+    user_language: str | None = None
     workspace: Workspace
     current_user: AppUser
     show_feedback_buttons: bool = False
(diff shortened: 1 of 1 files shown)


daras_ai_v2/bots.py
```diff
@@ -133,15 +131,8 @@ def __init__(self):
...
-            self.workspace = self.bi.workspace
-            self.current_user = self.bi.created_by
-        else:
-            # TODO: remove this once all bots have been migrated to workspaces
-            self.workspace = Workspace.objects.get_or_create_from_uid(
-                self.bi.billing_account_uid
-            )[0]
-            self.current_user = self.workspace.created_by
+        self.workspace = self.bi.workspace
+        self.current_user = self.bi.created_by
```
nikochiko, line 135: no longer needed.

i confirmed that there are no bot integrations that do not have a workspace

nikochiko:
test failure is due to google translate

nikochiko:
we need more code to fix this in the case of web chatbots. it doesn't error in 500 anymore but sends an empty 200 chunk and then 404 without the connected message (although it does reach that part and enter `bot.send_msg(PAGE_NOT_CONNECTED_ERROR)` code).

this does fix it for other integrations though and we should merge this first because it improves the current state of the codebase. i'll make another PR to fix the web integration case

devxpy:
I think if we convert `bot.send_msg(PAGE_NOT_CONNECTED_ERROR)` to `raise UserError(PAGE_NOT_CONNECTED_ERROR)`, it will fix the web issue too?

devxpy (approved):

nikochiko:
> I think if we convert bot.send_msg(PAGE_NOT_CONNECTED_ERROR) to raise UserError(PAGE_NOT_CONNECTED_ERROR), it will fix the web issue too?

this didn't work

PR #558, opened 2024-12-06
Add support for o1-preview & o1-mini

nikochiko:
- **add pricing entry for o1-preview & o1-mini**
- **allow temperature=None, LLMSpec.supports_temperature, and use max_completion_tokens instead of max_tokens**

### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/language_model.py
+++ b/daras_ai_v2/language_model.py
@@ -71,10 +71,35 @@ class LLMSpec(typing.NamedTuple):
     is_vision_model: bool = False
     is_deprecated: bool = False
     supports_json: bool = False
+    supports_temperature: bool = True
 
 
 class LargeLanguageModels(Enum):
-    # https://platform.openai.com/docs/models/gpt-4o
+    # https://platform.openai.com/docs/models#o1
+    o1_preview = LLMSpec(
+        label="o1-preview (openai)",
+        model_id="o1-preview-2024-09-12",
+        llm_api=LLMApis.openai,
+        context_window=128_000,
+        price=50,
+        is_vision_model=False,
+        supports_json=False,
+        supports_temperature=False,
+    )
+
+    # https://platform.openai.com/docs/models#o1
+    o1_mini = LLMSpec(
+        label="o1-mini (openai)",
+        model_id="o1-mini-2024-09-12",
+        llm_api=LLMApis.openai,
+        context_window=128_000,
+        price=13,
+        is_vision_model=False,
+        supports_json=False,
+        supports_temperature=False,
+    )
+
+    # https://platform.openai.com/docs/models#gpt-4o
     gpt_4_o = LLMSpec(
         label="GPT-4o (openai)",
         model_id="gpt-4o-2024-08-06",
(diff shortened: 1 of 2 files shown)


daras_ai_v2/language_model.py
```diff
@@ -71,10 +71,35 @@ class LLMSpec(typing.NamedTuple):
...
+    # https://platform.openai.com/docs/models#o1
+    o1_preview = LLMSpec(
+        label="o1-preview (openai)",
+        model_id="o1-preview-2024-09-12",
+        llm_api=LLMApis.openai,
+        context_window=128_000,
+        price=50,
+        is_vision_model=False,
+        supports_json=False,
+        supports_temperature=False,
```
nikochiko, line 87: https://platform.openai.com/docs/guides/reasoning#beta-limitations

nikochiko: we should get rid of this once out of beta

daras_ai_v2/language_model.py
```diff
@@ -743,7 +771,7 @@ def _run_chat_model(
             return _run_openai_chat(
                 model=model,
                 avoid_repetition=avoid_repetition,
-                max_tokens=max_tokens,
+                max_completion_tokens=max_tokens,
```
nikochiko, line 774: https://platform.openai.com/docs/guides/reasoning#controlling-costs

pyproject.toml
```diff
@@ -41,7 +41,7 @@ pdftotext = "^2.2.2"
 "pdfminer.six" = "^20221105"
 google-api-python-client = "^2.80.0"
 oauth2client = "^4.1.3"
-tiktoken = "^0.7.0"
+tiktoken = "^0.8"
```
nikochiko, line 44: because tiktoken 0.7 didn't support o1 tokeniser lookup

daras_ai_v2/language_model.py
nikochiko: it has been deprecated completely (errors with max_tokens). + max_tokens = max_completion_tokens for models that are not o1

```
To manage costs with the o1 series models, you can limit the total number of tokens the model generates (including both reasoning and completion tokens) by using the [max_completion_tokens](https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_completion_tokens) parameter.

In previous models, the max_tokens parameter controlled both the number of tokens generated and the number of tokens visible to the user, which were always equal. However, with the o1 series, the total tokens generated can exceed the number of visible tokens due to the internal reasoning tokens.

Because some applications might rely on max_tokens matching the number of tokens received from the API, the o1 series introduces max_completion_tokens to explicitly control the total number of tokens generated by the model, including both reasoning and visible completion tokens. This explicit opt-in ensures no existing applications break when using the new models. The max_tokens parameter continues to function as before for all previous models.
```

nikochiko: https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_tokens

devxpy (approved):

PR #572, opened 2024-12-20
support all non exportable google workspace docs

milovate:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/gdrive_downloader.py
+++ b/daras_ai_v2/gdrive_downloader.py
@@ -1,5 +1,5 @@
 import io
-
+import mimetypes
 from furl import furl
 import requests
 
@@ -73,8 +73,13 @@ def gdrive_download(f: furl, mime_type: str, export_links: dict) -> tuple[bytes,
 
     if f.host != "drive.google.com":
         # export google docs to appropriate type
-        export_mime_type, _ = docs_export_mimetype(f)
-        if f_url_export := export_links.get(export_mime_type, None):
+        export_mime_type, _, is_google_workspace_doc = docs_export_mimetype(
+            f, mime_type
+        )
+
+        if is_google_workspace_doc and (
+            f_url_export := export_links.get(export_mime_type, None)
+        ):
             r = requests.get(f_url_export)
             file_bytes = r.content
             raise_for_status(r, is_user_url=True)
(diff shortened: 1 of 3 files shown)


devxpy (commented):
I think we dont need the docs_export_mimetype we can just have a conversion dict that maps  the supported_mimetypes to their respective export mime type.

Then just lookup the available mimetypes in exportLinks

daras_ai_v2/gdrive_downloader.py
```diff
@@ -71,9 +73,16 @@ def gdrive_download(f: furl, mime_type: str, export_links: dict) -> tuple[bytes,
     # get metadata
     service = discovery.build("drive", "v3")
 
+    docs_export_mimetype = {
```
devxpy, line 76: make this a top level const please

devxpy (approved):

daras_ai_v2/gdrive_downloader.py
```diff
@@ -62,7 +62,9 @@ def gdrive_list_urls_of_files_in_folder(f: furl, max_depth: int = 4) -> list[str
     return filter(None, urls)
 
 
-def gdrive_download(f: furl, mime_type: str, export_links: dict) -> tuple[bytes, str]:
+def gdrive_download(
+    f: furl, mime_type: str, export_links: dict = {}
```
devxpy, line 66: image

devxpy (approved):

devxpy:
update type definition in `FileMetadata` and remove __init__

```py
class FileMetadata(models.Model):
    name = models.TextField(default="", blank=True)
    etag = models.CharField(max_length=255, null=True)
    mime_type = models.CharField(max_length=255, default="", blank=True)
    total_bytes = models.PositiveIntegerField(default=0, blank=True)

    export_links: dict[str, str] | None = None
```

daras_ai_v2/gdrive_downloader.py
```diff
@@ -1,12 +1,19 @@
...
-
+import typing
 from furl import furl
 import requests
 
 from daras_ai_v2.exceptions import UserError
 from daras_ai_v2.functional import flatmap_parallel
 from daras_ai_v2.exceptions import raise_for_status
 
+docs_export_mimetype = {
```
devxpy, line 10: `DOCS_EXPORT_MIMETYPES`

devxpy:
LGTM

PR #585, opened 2025-01-08
move handles from appuser to workspace (with fallback)

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


handles/models.py
```diff
@@ -118,14 +119,25 @@ class Meta:
...
 
-    def clean(self):
+    def _validate_exclusive(self):
+        if (
+            self.has_workspace
+            and self.has_user
+            and self.workspace.created_by_id == self.user.id
+        ):
+            # TODO: remove this once all handles are migrated
+            return
```
nikochiko, line 129: to handle an in-between state where the same handle can be attached to a personal workspace & its user. won't be needed after `cleanup_handles()` is run in the script

devxpy:
missed?
https://github.com/GooeyAI/gooey-server/blob/bd355fab98f67350ef6585fd96c9c7583eb3c3eb/workspaces/views.py#L468

routers/root.py
```diff
@@ -643,10 +643,17 @@ def recipe_or_handle_or_static(
...
+        user = handle.user
+    else:
+        user = None
+
+    if user:
         with page_wrapper(request):
-            user_profile_page(request, handle.user)
-        return dict(meta=get_meta_tags_for_profile(handle.user))
+            user_profile_page(request, user)
+        return dict(meta=get_meta_tags_for_profile(user))
```
devxpy, line 656: pass handle into get_meta_tags_for_profile directly instead of doing a 2x lookup

routers/account.py
```diff
@@ -329,11 +329,11 @@ def _render_run(pr: PublishedRun):
         return
 
     if workspace.is_personal:
-        if request.user.handle:
+        if handle := request.user.get_handle():
```
devxpy, line 332: this should be workspace.handle

nikochiko: done

routers/root.py
nikochiko: done

nikochiko:
thanks, fixed.

devxpy (approved):

app_users/models.py
```diff
@@ -248,6 +250,12 @@ def cached_workspaces(self) -> list["Workspace"]:
             ).order_by("-is_personal", "-created_at")
         ) or [self.get_or_create_personal_workspace()[0]]
 
+    def get_handle(self) -> Handle | None:
+        if self.handle:
+            return self.handle
+        workspace, _ = self.get_or_create_personal_workspace()
+        return workspace.handle
```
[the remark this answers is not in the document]
devxpy: Yeah sounds good

PR #593, opened 2025-01-16
show last ran time instead of updated_at always

anish-work:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/base.py
+++ b/daras_ai_v2/base.py
@@ -358,11 +358,7 @@ class BasePage:
                     with gui.nav_item(url, active=tab == self.tab):
                         gui.html(tab.title)
 
-            if (
-                self.current_pr
-                and not self.current_pr.is_root()
-                and self.tab == RecipeTabs.run
-            ):
+            if self.current_pr and self.tab == RecipeTabs.run:
                 with gui.div(
                     className="container-margin-reset d-none d-md-block",
                     style=dict(
@@ -490,8 +486,17 @@ class BasePage:
             gui.html("Unpublished changes")
 
     def _render_saved_timestamp(self, pr: PublishedRun):
+        last_run_at = gui.session_state.get(
+            StateKeys.updated_at, datetime.datetime.today()
+        )
+
+        if isinstance(last_run_at, str):
+            last_run_at = datetime.datetime.fromisoformat(last_run_at)
+
         with gui.tag("span", className="text-muted"):
-            gui.write(f"{get_relative_time(pr.updated_at)}")
+            gui.write(
+                f"{get_relative_time(pr.updated_at if pr.is_root() else last_run_at)}"
+            )
 
     def _render_options_button_with_dialog(self):
         ref = gui.use_alert_dialog(key="options-modal")
--- a/daras_ai_v2/base.py
+++ b/daras_ai_v2/base.py
@@ -358,7 +358,7 @@ class BasePage:
                     with gui.nav_item(url, active=tab == self.tab):
                         gui.html(tab.title)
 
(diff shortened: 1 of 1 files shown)


daras_ai_v2/base.py
```diff
@@ -490,8 +486,17 @@ def _render_unpublished_changes_indicator(self):
...
+            StateKeys.updated_at, datetime.datetime.today()
+        )
+
+        if isinstance(last_run_at, str):
+            last_run_at = datetime.datetime.fromisoformat(last_run_at)
+
         with gui.tag("span", className="text-muted"):
-            gui.write(f"{get_relative_time(pr.updated_at)}")
+            gui.write(
+                f"{get_relative_time(pr.updated_at if pr.is_root() else last_run_at)}"
```
devxpy, line 498: you wanna show the published run last updated_at time on the published run root, and show the saved run updated_at time on a saved run

daras_ai_v2/base.py
```diff
@@ -358,11 +358,7 @@ def render(self):
                     with gui.nav_item(url, active=tab == self.tab):
                         gui.html(tab.title)
 
-            if (
-                self.current_pr
-                and not self.current_pr.is_root()
-                and self.tab == RecipeTabs.run
-            ):
+            if self.current_pr and self.tab == RecipeTabs.run:
                 with gui.div(
```
devxpy, line 362: probably push this div inside render_saved_timestamp?

daras_ai_v2/base.py
```diff
@@ -490,8 +486,17 @@ def _render_unpublished_changes_indicator(self):
             gui.html("Unpublished changes")
 
     def _render_saved_timestamp(self, pr: PublishedRun):
```
devxpy, line 488: use the current_sr / current_pr here instead of passing down the pr and using statekeys.updated_at

daras_ai_v2/base.py
anish-work: There is specific styling to this div, which is relative to the parent. We won't need to always absolutely place the saved timestamp. imo @devxpy

devxpy: ok - in that case you should pull the component out of this class too!

anish-work: @devxpy I moved the div inside the fn, after lookig at other function I think you are right. Apologies for the dealy

daras_ai_v2/base.py
```diff
@@ -489,9 +475,22 @@ def _render_unpublished_changes_indicator(self):
...
+                    top="50%",
+                    right="0",
+                    transform="translateY(-50%)",
+                ),
+            ):
+                show_last_saved = pr.saved_run == sr or pr.is_root()
+                with gui.tag("span", className="text-muted"):
+                    gui.write(
+                        f"{get_relative_time(pr.updated_at if show_last_saved else sr.updated_at)}"
+                    )
```
devxpy, line 493:
```py
def _render_saved_timestamp(self):
    if not (self.current_pr and self.tab == RecipeTabs.run):
		return
    if pr.saved_run_id == sr.id or pr.is_root():
		dt = pr.updated_at
	else:
		dt = sr.updated_at
    with gui.div(
        className="container-margin-reset d-none d-md-block",
        style=dict(
            position="absolute",
            top="50%",
            right="0",
            transform="translateY(-50%)",
        ),
    ), gui.tag("span", className="text-muted"):
        gui.write(get_relative_time(dt))
```

devxpy (approved):

PR #595, opened 2025-01-22
Add support for scheduling daily message export function for integrations

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


recipes/VideoBotsStats.py
```diff
@@ -309,14 +349,105 @@ def render(self):
...
+                        (function() {{
+                            let el = document.createElement("a");
+                            el.setAttribute("href", "data:file/csv;base64,{b64}");
+                            el.setAttribute("download", "{slugify(bi.name)}.csv");
+                            document.body.appendChild(el);
+                            el.click();
+                            document.body.removeChild(el);
+                        }})();
+                        """
+                    )
```
nikochiko, line 365: this lets the server handle the effect after click - so that we don't prepare the CSV file on every page load

recipes/VideoBotsStats.py
```diff
@@ -394,13 +524,17 @@ def render_date_view_inputs(self, bi):
         return start_date, end_date, view, factor, trunc_fn
 
 
-def calculate_overall_stats(*, bi, run_title, run_url):
+def get_conversations_and_messages(bi) -> tuple[ConversationQuerySet, MessageQuerySet]:
     conversations: ConversationQuerySet = Conversation.objects.filter(
         bot_integration=bi
     ).order_by()  # type: ignore
+    messages: MessageQuerySet = Message.objects.filter(conversation__in=conversations).order_by()  # type: ignore
+    return conversations, messages
```
nikochiko, line 532: to make this step reusable in the UI & `run_all_scheduled_functions` script

scripts/run_all_scheduled_functions.py
```diff
@@ -0,0 +1,5 @@
+from bots.tasks import run_all_scheduled_functions
+
+
+def run():
+    run_all_scheduled_functions.delay()
```
nikochiko, line 5: this file isn't needed anymore with `celery beat`, but can be used to trigger running all functions manually

bots/tasks.py
```diff
@@ -211,6 +216,49 @@ def send_broadcast_msg(
         # save_broadcast_message(convo, text, msg_id)
 
 
+@shared_task
+def run_all_scheduled_functions():
+    for sf in BotIntegrationScheduledFunction.objects.select_related(
+        "bot_integration"
+    ).exclude(last_run_at__gt=timezone.now() - timedelta(hours=23)):
```
nikochiko, line 223: @devxpy i had to make this `timedelta(hours=23)` instead of `days=1` because the tasks will be started after `H:00` on day 1, and we want to re-run those that ran after `H:00` and before `H+1:00`.

PR #600, opened 2025-01-30
ZOHO - Transaction sync script

anish-work:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


devxpy:
use the django extensions runscript format for the script https://django-extensions.readthedocs.io/en/latest/runscript.html

devxpy:
I also don't see any use for the classes here, I think it would be better to just use functions. The logger you can just use loguru like we do everywhere else in code. The mapping_config & bulk_uploader you're only initializing once, so it makes sense to keep it in a global scope and pass it around instead of storing it in class.

scripts/zoho_crm.py
```diff
@@ -0,0 +1,583 @@
...
+        self,
+        start_date: Optional[datetime] = None,
+        end_date: Optional[datetime] = None,
+        positive_only: bool = True,
+        limit: int | None = None,
+        dry_run: bool = False,
+    ):
+        stats = {"processed": 0, "successful": 0, "failed": 0, "errors": []}
+        # Build transaction query with optional date filtering
+        transaction_query = AppUserTransaction.objects.all().order_by("created_at")
```
nikochiko, line 469: `AppUserTransaction.objects.select_related("user", "workspace__created_by").all().order_by(...)`

(fewer database queries)

devxpy:
bump

scripts/zoho_crm.py
```diff
@@ -0,0 +1,583 @@
...
+
+    def _load_mapping_config(self) -> Dict:
+        """
+        :return: Mapping configuration dictionary
+        """
+        default_config = {
+            "contact_mapping": {
+                "Gooey_User_ID": {"db_key": "uid"},
+                "Gooey_Admin_Link": {
+                    "db_key": "id",
```
devxpy, line 133: use named tuple instead of dict

scripts/zoho_crm.py
```diff
@@ -0,0 +1,583 @@
...
+        """
+        self.logger = logging.getLogger(self.__class__.__name__)
+        self.mapping_config = self._load_mapping_config()
+
+    def _load_mapping_config(self) -> Dict:
+        """
+        :return: Mapping configuration dictionary
+        """
+        default_config = {
+            "contact_mapping": {
```
devxpy, line 130: contact_mapping, transaction_mapping & workspace_mapping can be 3 top level constants instead of a dynamic string inside a dict generated from a class method which is very un-necessary and adds cruft

scripts/zoho_crm.py
```diff
@@ -0,0 +1,583 @@
...
+# Set up Django environment
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "daras_ai_v2.settings")
+django.setup()
+
+from app_users.models import AppUserTransaction, PaymentProvider, TransactionReason
+from daras_ai_v2 import settings
+
+ZOHO_CONTACT_API = "https://www.zohoapis.com/crm/v2/Contacts"
+ZOHO_DEAL_API = "https://www.zohoapis.com/crm/v7/Deals"
+ZOHO_HEADERS = {"Authorization": f"Bearer {settings.ZOHO_AUTH_CODE}"}
```
devxpy, line 30: generate this inside a method like we do in `azure_image_moderation` so that settings can be loaded at runtime if needed

scripts/zoho_crm.py
```diff
@@ -0,0 +1,583 @@
...
+ZOHO_BULK_FILE_UPLOAD_API = "https://content.zohoapis.com/crm/v7/upload"
+ZOHO_BULK_CREATE_JOB = "https://www.zohoapis.com/crm/bulk/v7/write"
+ZOHO_ORG_ID = settings.ZOHO_ORG_ID
```
devxpy, line 33: un-necessary to re-declare this variable

scripts/zoho_crm.py
```diff
@@ -0,0 +1,583 @@
...
+
+                # Apply transformation if specified
+                if value is not None:
+                    zoho_fields[zoho_field] = (
+                        transformer(value) if transformer else value
+                    )
+                else:
+                    zoho_fields[zoho_field] = field_config.get("default") or "None"
+
+            except Exception as e:
```
devxpy, line 261: https://peps.python.org/pep-0008/#programming-recommendations

image

scripts/zoho_crm.py
```diff
@@ -0,0 +1,583 @@
...
+                "Updated_At": {
+                    "db_key": "updated_at",
+                    "transformer": lambda date: date.strftime("%Y-%m-%d"),
+                },
+            },
+        }
+
+        return default_config
+
+    def _deep_merge(self, base: Dict, update: Dict):
```
devxpy, line 223: unused method

PR #604, opened 2025-02-06
update whatsapp intergration decription text

anish-work:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

### DIFF
--- a/recipes/VideoBots.py
+++ b/recipes/VideoBots.py
@@ -1779,7 +1779,7 @@ connect_choices = [
     ConnectChoice(
         platform=Platform.WHATSAPP,
         img="https://storage.googleapis.com/dara-c1b52.appspot.com/daras_ai/media/1e49ad50-d6c9-11ee-99c3-02420a000123/thumbs/Digital_Inline_Green_400x400.png",
-        label="[Read our guide](https://docs.gooey.ai/guides/how-to-deploy-an-ai-copilot/deploy-on-whatsapp) to connect your own mobile # (that's not already on WhatsApp) or [upgrade](https://gooey.ai/pricing) for a number on us.Read our guide to connect your own mobile # (that's not already on WhatsApp) or upgrade for a number on us.",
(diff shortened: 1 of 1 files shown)


recipes/VideoBots.py
```diff
@@ -1776,7 +1776,7 @@ class ConnectChoice(typing.NamedTuple):
     ConnectChoice(
         platform=Platform.WHATSAPP,
         img="https://storage.googleapis.com/dara-c1b52.appspot.com/daras_ai/media/1e49ad50-d6c9-11ee-99c3-02420a000123/thumbs/Digital_Inline_Green_400x400.png",
-        label="Connect your own new mobile # (that's not already on WhatsApp) or [upgrade to Business](https://gooey.ai/pricing) for a number on us.",
+        label="[Read our guide](https://docs.gooey.ai/guides/how-to-deploy-an-ai-copilot/deploy-on-whatsapp) to connect your own mobile # (that's not already on WhatsApp) or [upgrade](https://gooey.ai/pricing) for a number on us.Read our guide to connect your own mobile # (that's not already on WhatsApp) or upgrade for a number on us.",
```
devxpy, line 1779: why is this repeated

recipes/VideoBots.py
```diff
@@ -1776,7 +1776,7 @@ class ConnectChoice(typing.NamedTuple):
     ConnectChoice(
         platform=Platform.WHATSAPP,
         img="https://storage.googleapis.com/dara-c1b52.appspot.com/daras_ai/media/1e49ad50-d6c9-11ee-99c3-02420a000123/thumbs/Digital_Inline_Green_400x400.png",
-        label="Connect your own new mobile # (that's not already on WhatsApp) or [upgrade to Business](https://gooey.ai/pricing) for a number on us.",
+        label="[Read our guide](https://docs.gooey.ai/guides/how-to-deploy-an-ai-copilot/deploy-on-whatsapp) to connect your own mobile # (that's not already on WhatsApp) or [upgrade](https://gooey.ai/pricing) for a number on us.",
```
devxpy, line 1779: umm, i think the https://gooey.ai/pricing should actually be https://gooey.ai/account/billing/ ?

devxpy (approved):

PR #607, opened 2025-02-10
fix: set upper bound on max_tokens parameter for LLMs

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/language_model_settings_widgets.py
+++ b/daras_ai_v2/language_model_settings_widgets.py
@@ -49,17 +49,16 @@ def language_model_settings(selected_models: str | list[str] | None = None) -> N
     with col1:
         gui.checkbox("Avoid Repetition", key="avoid_repetition")
 
-    if any(map(lambda llm: llm.supports_json, llms)):
-        with col2:
-            gui.selectbox(
-                f"###### {field_title_desc(LanguageModelSettings, 'response_format_type')}",
-                options=[None, "json_object"],
-                key="response_format_type",
-                format_func={
-                    None: BLANK_OPTION,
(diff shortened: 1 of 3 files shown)


devxpy:
reminder to update this for all LLMs with defaults

daras_ai_v2/language_model_settings_widgets.py
```diff
@@ -32,16 +32,24 @@ def language_model_selector(
...
+            llms.append(LargeLanguageModels[model])
+        except KeyError:
+            pass
 
     col1, col2 = gui.columns(2)
     with col1:
         gui.checkbox("Avoid Repetition", key="avoid_repetition")
-    if not llm or llm.supports_json:
+
+    if any(map(lambda llm: llm.supports_json, llms)):
```
devxpy, line 52: we support json output artificially via prompting for models that don't support it natively, so we might wanna leave this like we had earlier?

daras_ai_v2/language_model_settings_widgets.py
```diff
@@ -87,7 +105,9 @@ def language_model_settings(selected_model: str = None):
             min_value=1,
             max_value=4,
         )
-    if llm and not llm.is_chat_model and llm.llm_api == LLMApis.openai:
+    if llms and any(
```
devxpy, line 108: would prefer `any(<cond> for item in llms)` instead of the lambda version here!

devxpy (approved):

PR #609, opened 2025-02-14
Saved list layout

anish-work:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


daras_ai_v2/base.py
```diff
@@ -2276,43 +2280,255 @@ def render_published_run_preview(self, published_run: PublishedRun):
...
-        self,
-        *,
-        published_run: PublishedRun,
-        allow_hide: bool,
-    ):
+    def render_published_run_full_width(self, published_run: PublishedRun, **kwargs):
+        hide_author_column = True if "hide_author_column" in kwargs else False
+        use_workspace_author = True if "use_workspace_author" in kwargs else False
+        max_desc_words = 230 if hide_author_column else 180
+        max_desc_words_mobile = 100
```
nikochiko, line 2287: Instead of this, you can use arguments with default values. That is more pythonic, clearer.

```python
def render_published_run_full_width(self, published_run: PublishedRun, show_author: bool = False, show_workspace: bool = False, show_workflow: bool = False, show_visibility: bool = False):
    max_description_words = 230 if hide_author_column else 100
    if show_author:
        ...
    if show_workflow:
        ...
    ....
```

daras_ai_v2/base.py
```diff
@@ -2276,43 +2280,255 @@ def render_published_run_preview(self, published_run: PublishedRun):
...
+
+                                        gui.write(
+                                            f"#### {truncate_text_words(tb.h1_title, 50)}",
+                                        )
+                                with gui.div(
+                                    className="d-md-flex d-none align-items-center mb-1 ms-2",
+                                    style={"font-size": "0.9rem"},
+                                ):
+                                    for pill in pills:
+                                        pill()
```
nikochiko, line 2393: how about...

```python
if show_workflow:
    workflow = Workflow(published_run.workflow)
    gui.pill(workflow.short_title, className="...")
if show_visibility: 
    visibility = PublishedRunVisibility(published_run.visibility)
    gui.pill(visibility.get_badge_html(), unsafe_allow_html=True, className="...")
```

nikochiko (dismissed):

daras_ai_v2/base.py
```diff
@@ -2276,43 +2280,255 @@ def render_published_run_preview(self, published_run: PublishedRun):
...
+                            responsive=False,
+                            show_as_link=False,
+                        )
+                    )
+                    if version.change_notes:
+                        with gui.div(className="mt-2"):
+                            gui.caption(
+                                f"{icons.notes} {html.escape(truncate_text_words(version.change_notes, 20))}",
+                                unsafe_allow_html=True,
+                            )
```
nikochiko, line 2439: please break down chunks with their own responsibility into their own functions, so it can be repeated / easier to understand. e.g.

```python
_render_author_column(
    published_run=published_run,
    change_notes=change_notes,
    show_workspace=show_workspace,
)
```

daras_ai_v2/base.py
```diff
@@ -2276,43 +2280,255 @@ def render_published_run_preview(self, published_run: PublishedRun):
...
+                                )
+                                if not use_workspace_author
+                                else self.render_workspace_author(
+                                    published_run.workspace,
+                                    image_size="24px",
+                                    text_size="0.9rem",
+                                    responsive=False,
+                                    show_as_link=False,
+                                )
+                            )
```
nikochiko, line 2507: This would be easier to read with proper `if-else` conditions. 

```python
if not use_workspace_author:
   self.render_author(...)
else:
   self.render_workspace_author(...)
```

Please use the same elsewhere when it spans multiple lines

devxpy (approved):

PR #619, opened 2025-03-07
Refactor format_chat_entry to add input_urls for images, audio, and d…

milovate:
…ocuments to content_text

### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


devxpy (approved):

daras_ai_v2/language_model.py
```diff
@@ -1964,11 +1964,27 @@ def entry_to_prompt_str(entry: ConversationEntry) -> str:
...
+    content_text: str,
+    input_images: typing.Optional[list[str]] = None,
+    input_audio: typing.Optional[str] = None,
+    input_documents: typing.Optional[list[str]] = None,
+    render_input_urls: typing.Optional[bool] = False,
 ) -> ConversationEntry:
-    if images:
+
+    input_urls = []
+    if input_images and not render_input_urls:
```
devxpy: this condition didn't make a lot of sense. the parameter is called render_input_urls, but the function then goes on to do the exact opposite (i.e. NOT render URLs)? better naming please

daras_ai_v2/language_model.py
```diff
@@ -1964,11 +1964,27 @@ def entry_to_prompt_str(entry: ConversationEntry) -> str:
...
+        input_urls.append(f"Image URLs: {', '.join(input_images)}")
+    if input_audio and not render_input_urls:
+        input_urls.append(f"Audio URL: {input_audio}")
+    if input_documents and not render_input_urls:
+        input_urls.append(f"Document URLs: {', '.join(input_documents)}")
+
+    content = content_text + ("\n" + "\n".join(input_urls) if input_urls else "")
```
devxpy: you don't need to do the multiple concatenation and then a join() + an inline if-condition. it could have all been folded into one join()

PR #637, opened 2025-04-15
format with ruff

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


devxpy:
Thanks much needed but I'm going to hold this off until a good point where other PRs are merged!

devxpy:
go now

nikochiko:
so the full package (linter + formatter) makes more changes. the earlier changes were only linter.

please take a look at these are some code changes i made to make linter pass: https://github.com/GooeyAI/gooey-server/pull/637/files/587a7fc681f9b02aa55f9d29639d173c52a2a5ed..4c84ae67fb908d400170ffdb463f15306c31ed59
- removed `_reflect(...)` and `_add_shadow(...)` from image segmentation because they weren't used anywhere in the code
- removed `page = VideoBotsStatsPage()` in `bots/tests.py` because it was unused. i can revert if this was intentional and part of the tests
- removing other unused variables

more changes are due to the formatter. i added so we have consistent styles - and ignored some rules to match our codebase.
only the linter is also good and does most of what we needed `black` for. i don't mind a few inconsistencies in style if we don't use the formatter.

handles/models.py
```diff
@@ -246,7 +245,7 @@ def _generate_handle_options_for_team_workspace(
     handle_name = _make_handle_from(display_name)
     yield handle_name[:HANDLE_MAX_LENGTH]
     for i in range(1, 10):
-        yield f"{handle_name[:HANDLE_MAX_LENGTH-1]}{i}"
+        yield f"{handle_name[: HANDLE_MAX_LENGTH - 1]}{i}"
```
nikochiko, line 248: this is the only change where i don't like ruff's suggestion

nikochiko:
note: https://github.com/GooeyAI/gooey-server/pull/640 needs to be merged before this PR because `ModelTrainer` right now depends on one of black's dependencies :p 

the tests still pass because i think github's default python image comes with black & its dependencies.

`poetry run python -X importtime 'server'` failed locally is when i realised

devxpy (approved):

devxpy (approved):

PR #639, opened 2025-04-17
deprecate LLMs with redirect logic

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/language_model.py
+++ b/daras_ai_v2/language_model.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import base64
 import json
 import mimetypes
@@ -80,10 +82,14 @@ class LLMSpec(typing.NamedTuple):
     price: int = 1
     is_chat_model: bool = True
     is_vision_model: bool = False
-    is_deprecated: bool = False
     supports_json: bool = False
     supports_temperature: bool = True
     is_audio_model: bool = False
+    is_deprecated: bool = False
+    # redirect to a different LLM when deprecated
+    redirect_to: typing.Literal[tuple(m.name for m in LargeLanguageModels)] | None = (
+        None
+    )
 
 
 class LargeLanguageModels(Enum):
@@ -196,17 +202,19 @@ class LargeLanguageModels(Enum):
     )
 
     chatgpt_4_o = LLMSpec(
-        label="ChatGPT-4o (openai) 🧪",
+        label="ChatGPT-4o [Swapped by GPT-4o] (openai) 🧪",
         model_id="chatgpt-4o-latest",
         llm_api=LLMApis.openai,
         context_window=128_000,
         max_output_tokens=16_384,
         price=10,
         is_vision_model=True,
+        is_deprecated=True,
+        redirect_to="gpt_4_o",
     )
     # https://platform.openai.com/docs/models/gpt-4-turbo-and-gpt-4
     gpt_4_turbo_vision = LLMSpec(
-        label="GPT-4 Turbo with Vision (openai)",
+        label="GPT-4 Turbo with Vision [Swapped by GPT-4o] (openai)",
         model_id=(
             "openai-gpt-4-turbo-2024-04-09-prod-eastus2-1",
             "gpt-4-turbo-2024-04-09",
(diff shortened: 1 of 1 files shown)


daras_ai_v2/language_model.py
```diff
@@ -819,6 +874,13 @@ def run_language_model(
     ), "Pleave provide exactly one of { prompt, messages }"
 
     model: LargeLanguageModels = LargeLanguageModels[str(model)]
+
+    if model.is_deprecated:
+        if model.redirect_to:
+            return run_language_model(**locals() | {"model": model.redirect_to})
+        else:
+            raise UserError(f"Model {model} is deprecated.")
```
nikochiko, line 882: this is new behavior for deprecated models without a `redirect_to` option. raising `UserError` here.

earlier, the behavior was to attempt to run the model regardless.

daras_ai_v2/language_model.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
nikochiko, line 1: needed for future reference to `LargeLanguageModels` for the type of `redirect_to` attribute.

devxpy:
I wonder if a more robust way to do this would be to define the fallback on the `LLMApis` level instead of the model level. because the fallbacks themselves will get deprecated eventually!

daras_ai_v2/language_model.py
```diff
@@ -819,6 +874,26 @@ def run_language_model(
     ), "Pleave provide exactly one of { prompt, messages }"
 
     model: LargeLanguageModels = LargeLanguageModels[str(model)]
+
+    if model.is_deprecated:
+        if model.redirect_to:
```
devxpy, line 879: why are these 2 nested conditions

nikochiko: the other way to write would be:
```python
if model.is_deprecated and model.redirect_to:
    ...
elif model.is_deprecated:
  ...
```

devxpy: sounds good but I dont see an elif condition required in your code

nikochiko: It is this at the moment:

```python
if model.is_deprecated:
  if model.redirect_to:
    return run_language_model(...)
  else:
    raise UserError("model is deprecated")
```

Yes moving to non-nested `if`-`elif` will be more obvious

devxpy: image

nikochiko: ah. that doesn't look right. will fix

nikochiko:
> I wonder if a more robust way to do this would be to define the fallback on the LLMApis level instead of the model level. because the fallbacks themselves will get deprecated eventually!

we can add `redirect_to` on those models and that chain will work?

nikochiko:
`run_language_model()` -> `run_language_model()` -> `run_language_model()`

devxpy:
@coderabbitai review

devxpy:
Also:
- the "[Redirects to GPT-4o-mini]" bit should be auto generated
- redirect_to can be a direct reference to the instance variable `gpt_4_o` instead of the string `"gpt_4_o"`

daras_ai_v2/language_model.py
devxpy: fyi: I would have dont it like this, but what you have works
```
if model.is_deprecated:
	if not model.redirect_to:
	    raise UserError(f"Model {model} is deprecated.")	
    return run_language_model(...)
```

nikochiko:
> redirect_to can be a direct reference to the instance variable gpt_4_o instead of the string "gpt_4_o"

this is what i did initially, but the way the `Enum` is written doesn't allow me to obtain the `name`. `.name` property is not available because within the class `gpt_4_o` refers to an instance of `LLMSpec` instance and not `LargeLanguageModels`. `__name__` also did not work.

devxpy (approved):

PR #652, opened 2025-04-28
update text on share pills

nikochiko:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


devxpy (dismissed):

widgets/saved_workflow.py
```diff
@@ -124,96 +118,110 @@ def render_saved_workflow_preview(
                                     style={"fontSize": "0.9rem"},
                                 )
                     with gui.div(
-                        className="d-none d-md-flex flex-1 align-items-center render_example_author_meta"
+                        className="d-none d-md-flex container-margin-reset align-items-center gap-2"
```
nikochiko, line 121: `flex-1` did not have any styles. `render_example_author_meta` is repeated in `render_saved_workflow_author`.

widgets/saved_workflow.py
```diff
@@ -124,96 +118,110 @@ def render_saved_workflow_preview(
...
+):
+    className = "d-flex align-items-center container-margin-reset gap-2 " + (
+        props.pop("className", "")
+    )
+    with gui.div(className=className, **props):
+        items = filter(None, items)
+        for i, item in enumerate(items):
+            if i != 0:
+                gui.write(separator)
+            item()
```
nikochiko, line 224: @devxpy thoughts on this abstraction?

it was getting tedious to put separators. it was also muddying the core logic in rendering functions. i'm not very happy about the `lambda`s though.

widgets/saved_workflow.py
```diff
@@ -124,96 +118,110 @@ def render_saved_workflow_preview(
...
+    hide_version_notes: bool = False,
+    hide_visibility_pill: bool = False,
+    **props,
+):
+    version = published_run.versions.latest()
+    items = []
+    if not hide_version_notes and version.change_notes:
+        items.append(
+            lambda: gui.caption(
+                f"{icons.notes} {html.escape(version.change_notes)}",
```
nikochiko, line 198: `html.escape` was missing here - allowed javascript xss.

widgets/saved_workflow.py
devxpy: just use css

devxpy (approved):

PR #698, opened 2025-06-12
show colored demo pill

anish-work:
### Q/A checklist

```bash
poetry lock --no-update
```
- [x] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate

---
  - https://app.asana.com/0/0/1210449894717252

### DIFF
--- a/widgets/saved_workflow.py
+++ b/widgets/saved_workflow.py
@@ -15,7 +15,7 @@ from daras_ai_v2.breadcrumbs import get_title_breadcrumbs
 from daras_ai_v2.meta_preview_url import meta_preview_url
 from daras_ai_v2.utils import get_relative_time
 from widgets.author import render_author_from_workspace, render_author_from_user
-from widgets.demo_button import get_demo_bots
+from widgets.demo_button import get_unique_demo_platforms
 
 if typing.TYPE_CHECKING:
     from daras_ai_v2.base import BasePage
@@ -88,7 +88,7 @@ def render_saved_workflow_preview(
 
 def render_title_pills(published_run: PublishedRun, workflow_pill: str | None):
     with gui.div(
-        className="d-md-flex d-none align-items-center ms-2 mt-1",
+        className="d-md-flex d-none align-items-center ms-2 mb-2",
         style={"font-size": "0.9rem"},
     ):
         if workflow_pill:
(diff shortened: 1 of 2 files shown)


widgets/saved_workflow.py
devxpy, line 1: The `platform_ids` returned by `get_demo_bots` should ideally be filtered at the SQL query level, not in Python. This will avoid unnecessary data transfer and improve performance, especially if the filtering logic becomes more complex or the dataset grows. Consider adding any filtering logic directly to the Django ORM query inside `get_demo_bots`.

devxpy, line 1: The `gui-pill-bg-color-{platform_id}` class is being added to a div, but there is no CSS or logic that actually uses this class for styling. The pill color is applied via inline styles in the parent `styled` context, not through this class. Unless you plan to add CSS that targets this class, it is currently unnecessary and can be removed.

widgets/saved_workflow.py
```diff
@@ -98,14 +98,37 @@ def render_title_pills(published_run: PublishedRun, workflow_pill: str | None):
...
+            rendered_ids.add(platform_id)
             platform = Platform(platform_id)
             label = f"{platform.get_icon()} {platform.get_title()}"
-            gui.pill(
-                label,
-                unsafe_allow_html=True,
-                className="border border-dark ms-2",
-            )
+            bg_color = platform.get_demo_button_color()
+            if not bg_color:
```
devxpy, line 110: This block can be simplified and improved:

- Instead of repeating the pill logic for colored and non-colored pills, just set the style dict conditionally and use a single `gui.tag` block.
- The current approach uses inline style dicts and bypasses the `gui.pill` component for colored pills, which is inconsistent. Ideally, the `pill` component should accept inline styles directly, or if it's just a thin wrapper, consider not using it at all and always use `gui.tag`.
- This will make the code more maintainable and avoid duplicating logic for rendering pills.

Example:
```python
if bg_color:
    style = dict(
        backgroundColor=bg_color,
        borderColor=bg_color,
        color="white",
    )
else:
    style = dict(
        backgroundColor="#f8f9fa",
        color="black",
    )
with gui.tag(
    "span",
    className="badge rounded-pill border ms-2",
    style=style,
):
    gui.html(label)
```

Consider refactoring the pill rendering logic as above and/or updating the `pill` component to support inline styles directly.

widgets/saved_workflow.py
```diff
@@ -98,14 +98,27 @@ def render_title_pills(published_run: PublishedRun, workflow_pill: str | None):
...
             label = f"{platform.get_icon()} {platform.get_title()}"
-            gui.pill(
-                label,
-                unsafe_allow_html=True,
-                className="border border-dark ms-2",
-            )
+            bg_color = platform.get_demo_button_color()
+            if bg_color:
+                style = dict(
+                    backgroundColor=bg_color + " !important",
```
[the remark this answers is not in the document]
anish-work: but we do need it for border-color as bootstrap overrides it using !important

devxpy (approved):

PR #706, opened 2025-06-17
Twilio Transfer call via functions

milovate:
### Q/A checklist

```bash
poetry lock --no-update
```
- [ ] Run tests 
```bash
```
```bash
```
```bash

```
```python
def my_function():
    import pandas as pd
    ...
```

### Legal Boilerplate


devxpy (commented):
Hi Milind,

The function `get_phone_number` is misleadingly named. According to [OpenAI's best practices for defining functions](https://platform.openai.com/docs/guides/function-calling?api-mode=chat#best-practices-for-defining-functions), function names should clearly describe their actual purpose. In this case, `get_phone_number` is not just retrieving a number for the LLM, but is actually responsible for transferring the call itself. This can cause confusion for future maintainers and makes the code less self-documenting. Please consider renaming the function to better reflect its true responsibility, such as `transfer_call` or similar.

Thanks!

daras_ai_v2/language_model_openai_realtime.py
```diff
@@ -212,106 +242,56 @@ def yield_from(gen: typing.Generator[typing.Any, None, T]) -> T:
...
+
+    try:
+        bi_id_decoded = api_hashids.decode(bi_id)[0]
+        bi = BotIntegration.objects.get(id=bi_id_decoded)
+    except (IndexError, BotIntegration.DoesNotExist) as e:
+        logger.debug(
+            f"could not find bot integration with bot_id={bi_id}, call_sid={call_sid} {e}"
+        )
+        capture_exception(e)
+        return None
```
devxpy, line 278: return a nice error message for the llm

devxpy (commented):
**Refactor suggestion for LLMTool usage:**

Currently, the code in `language_model_openai_audio.py` and `language_model_openai_realtime.py` uses a generic `LLMTool` type for the `tools` argument, and handles function calls/procedural logic for internal tools (like call transfer) directly. To improve maintainability and clarity:

- create a base class for `LLMTool` with type stubs for `name`, `spec`, `__call__` , etc.
- Create subclasses for each internal tool (such as call transfer)
- use this as the base for the generic `LLMTool`.
- Update the type stub for `run_language_model` so it accepts the new base class type for tools.

This will make it easier to extend and maintain the tool system, and clarify the contract for all LLM tools used in the codebase.

devxpy (approved):

PR #708, opened 2025-06-18
fix claude 4

nikochiko:
- **Revert "rollback using openai sdk for claude 4 models"**
- **fix claude to work with openai sdk: iter over response.__stream__() and not response**

---
  - https://app.asana.com/0/0/1210496356500391


devxpy:
Amazing!

devxpy (approved):

PR #746, opened 2025-07-30
feat: better version-restore logic by preferring to fetch pr from URL

nikochiko:
Currently, clicking on a version from a published-run page takes you to the individual saved-run page. The URL is like `https://gooey.ai/copilot/published-run-slug-pr_id_123/?run_id=...&uid=...`, so we can fetch the active published-run from the URL if it is there. This way, the flow of `{icon.save} Update` -> `{icon.save} Save` can be used to update the same published run.

It is not entirely consistent because versions can also contain updates to visibility settings or title/description. These changes are not rolled back with only the saved run. This can be fixed by changing the links and adding a param for the version ID, e.g. a URL such as `https://gooey.ai/copilot/published-run-slug-pr_id_123/version_id_123_here`.

---

- **refactor: stricter typing for extract_query_params**
- **feat: change sr-pr fetching logic to always prefer return pr from URL**

### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/base.py
+++ b/daras_ai_v2/base.py
@@ -1481,14 +1481,21 @@ class BasePage:
         cls, example_id: str, run_id: str, uid: str
     ) -> tuple[SavedRun, PublishedRun]:
         if run_id and uid:
-            sr = cls.get_sr_from_ids(run_id, uid)
-            pr = sr.parent_published_run() or cls.get_root_pr()
+            sr = cls.get_sr_from_ids(run_id=run_id, uid=uid)
+            pr = (
+                example_id
+                and cls.get_pr_from_example_id(example_id)
+                or sr.parent_published_run()
+                or cls.get_root_pr()
+            )
         else:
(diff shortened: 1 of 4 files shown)


daras_ai_v2/base.py
```diff
@@ -1481,14 +1481,21 @@ def get_sr_pr_from_query_params(
...
         else:
-            if example_id:
-                pr = cls.get_pr_from_example_id(example_id=example_id)
-            else:
-                pr = cls.get_root_pr()
+            pr = (
+                example_id
+                and cls.get_pr_from_example_id(example_id)
+                or cls.get_root_pr()
+            )
```
devxpy, line 1496: Hmm, I don't like these inline conditionals can we pls revert back to the multiline one

PR #870, opened 2026-01-12
Add team subscription plan & per-seat pricing

nikochiko:
### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate


daras_ai_v2/billing.py
```diff
@@ -412,84 +942,281 @@ def _render_upgrade_subscription_button(
...
-        workspace.subscription.cancel()
-        raise gui.RedirectException(
-            get_app_route_url(payment_processing_route), status_code=303
+    # Check if plan and seat selection are unchanged
+    current_seat_type = (
+        workspace.subscription and workspace.subscription.get_seat_type()
+    )
+    current_seats = (
+        workspace.subscription
+        and max(1, workspace.subscription.billed_seats().count())
```
nikochiko: max(1, ...) is unnecessary

devxpy:
i thought we had decided not to use workpsace level balance for per seat pricing and only use the workspace member level balance

daras_ai_v2/billing.py
```diff
@@ -1,34 +1,239 @@
...
+def _clear_pending_stripe_subscription_changes(
+    subscription: stripe.Subscription,
+) -> stripe.Subscription:
+    schedule = subscription.get("schedule")
+    schedule_id = schedule and (
+        isinstance(schedule, str)
+        and schedule
+        or isinstance(schedule, dict)
+        and schedule.get("id")
+        or None
```
devxpy:
```
if shchedule and insinstance(schedule, str):
   schedule_id = schedule
elif isinstance(schedule, dict):
   schedule_id = schedule.get("id")
else:
   schedule = None
```

daras_ai_v2/billing.py
```diff
@@ -89,166 +303,434 @@ def render_current_plan(workspace: "Workspace"):
...
-                    **render_local_date_attrs(
-                        next_invoice_ts,
-                        date_options={"day": "numeric", "month": "long"},
-                    ),
-                )
-
-        if plan is PricingPlan.ENTERPRISE:
-            # charge details are not relevant for Enterprise customers
-            return
+                ):
```
devxpy: revert back to single if-condition please

devxpy:
### Code review

Found 3 issues:

1. **Inverted guard in `WorkspaceInvite.accept()` blocks all non-TEAM/ENTERPRISE invites.** The condition `if PricingPlan.from_sub(...) not in [TEAM, ENTERPRISE]: raise ValidationError(...)` raises "workspace is full" for every plan that is *not* TEAM or ENTERPRISE (PRO, BUSINESS, CREATOR, legacy plans). The intent is to enforce seat limits only *for* TEAM/ENTERPRISE plans, but the condition is inverted — all other workspace types are permanently broken for invite acceptance.

https://github.com/GooeyAI/gooey-server/blob/13fd52f412f6814d65ee58bc9f284aab220af880/workspaces/models.py#L879-L884

2. **`pm.detach()` is unreachable — payment method left attached in Stripe.** In `_render_delete_payment_method_button`, `change_subscription(workspace, new_plan=PricingPlan.STARTER)` always raises `gui.RedirectException` internally, so the subsequent `pm.detach()` call never executes. The original code called `set_workspace_subscription` directly (which returns normally), then detached the payment method. After the refactor to use `change_subscription`, the detach is dead code.

https://github.com/GooeyAI/gooey-server/blob/13fd52f412f6814d65ee58bc9f284aab220af880/daras_ai_v2/billing.py#L1712-L1718

3. **`get_or_create_gooey_admin_seat_type()` omits the required `key` field.** `SeatType.key` is non-nullable, has no default, and is marked `editable=False`. The `get_or_create` call provides `name` as the lookup and several `defaults`, but never supplies `key`. On first creation (before `init_seat_types.py` has run), this will raise `IntegrityError`. This function is called at runtime from workspace signals, not just from the seeding script.

https://github.com/GooeyAI/gooey-server/blob/13fd52f412f6814d65ee58bc9f284aab220af880/workspaces/models.py#L673-L685

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

nikochiko:
#1 is intended behaviour. No other plan should have workspace members.

devxpy (approved):

devxpy (approved):

PR #909, opened 2026-03-09
Feature/gcs thumbnails

anish-work:
### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

---
  - https://app.asana.com/0/0/1213338365134503


anish-work:
@devxpy @nikochiko needs a review..

daras_ai_v2/meta_preview_url.py
```diff
@@ -21,21 +24,46 @@ def meta_preview_url(
...
+def _gcs_blob_exists(public_url: str) -> bool:
+    from daras_ai.image_input import gcs_bucket
+
+    f = furl(public_url)
+    # path segments: [bucket-name, ...blob-path-parts]
+    blob_name = "/".join(f.path.segments[1:])
+    try:
+        return gcs_bucket().blob(blob_name).exists()
+    except Exception:
+        return True  # fail open — assume exists to avoid hiding real thumbnails
```
nikochiko: @anish-work i think we should return `False` here to avoid showing a thumbnail that doesn't exist? we have a fallback to the `file_url`.

nikochiko:
@anish-work this doesn't make use of the preview image in history tab. i think that was also part of the task?

anish-work:
> @anish-work this doesn't make use of the preview image in history tab. i think that was also part of the task?

The commit got overwritten somehow, will add again

daras_ai_v2/base.py
```diff
@@ -2372,7 +2377,12 @@ def _render_run_preview(self, saved_run: SavedRun):
                 elif saved_run.error_msg:
                     gui.error(saved_run.error_msg, unsafe_allow_html=True)
                 else:
-                    self.render_run_preview_output(saved_run.to_dict())
+                    state = saved_run.to_dict()
+                    raw_url = self.preview_image(state)
+                    if raw_url:
+                        render_media_component(raw_url)
```
anish-work, line 2389: actually @nikochiko this part renders the render_media_component which uses the meta_preview_url

daras_ai_v2/meta_preview_url.py
anish-work: @nikochiko updated

daras_ai_v2/base.py
nikochiko: it still shows me the full URL on /video/history though. i think it's because Video outputs to `state["output_videos"]` which is not part of the keys that are checked by `preview_image`.

can you please add that too?

anish-work:
@nikochiko added

nikochiko (dismissed):
Thanks! LGTM

devxpy:
i think this will make page loads slow because of the network call by blob.exists() in the hot path of a page load

nikochiko:
`check_exists` is `False` by default

anish-work:
@devxpy This seems correct then ?

widgets/saved_workflow.py
```diff
@@ -309,21 +309,44 @@ def render_workflow_media(output_url: str, published_run: PublishedRun):
...
+):
+    """
+    Render an image that prefers a GCS thumbnail but falls back to the original URL
+    if the thumbnail doesn't exist.
+
+    Uses <object data=thumb><img src=original/></object>. The browser tries the thumb
+    first; on 404 it renders the inner <img>. This avoids a server-side GCS existence
+    check in the page-load hot path.
+    """
+    with gui.tag("object", data=thumb_url, style=style, className=className):
```
devxpy, line 362: if thumb_url == fallback_url, dont render object tag

devxpy:
going to /video/history seems to show a popup to download G9Ee1bvD83IBhc5o5jztQ_tmpgmvqb6rs.mp4

image

image

devxpy:
if you try making the thumb_url unreachable, the fallback images are oval instead of round

image

image

devxpy:
fal videos look weird
image

anish-work:
@devxpy Fixed the issue

anish-work:
> going to /video/history seems to show a popup to download G9Ee1bvD83IBhc5o5jztQ_tmpgmvqb6rs.mp4
> 
> image image

@devxpy not able to repro this

devxpy (commented):
A few findings from a fresh pass over the PR — one I'd flag as a real correctness regression, plus a behavior change worth confirming. Inline.

widgets/saved_workflow.py
```diff
@@ -309,21 +309,55 @@ def render_workflow_media(output_url: str, published_run: PublishedRun):
...
+            gui.html(
+                f'<video src={video_url!r} class="gui-video w-100" autoplay playsInline loop muted controls>'
+            )
+        else:
             gui.html(
                 f"""
-                <object data={preview_url!r} class="gui-video">
-                <video src={output_url!r} class="gui-video" autoplay playsInline loop muted>
+                <object data={preview_url!r} class="gui-video w-100">
+                <video src={video_url!r} class="gui-video" autoplay playsInline loop muted controls>
```
devxpy, line 327: **Critical: the fallback `<video>` plays the GIF thumbnail instead of the original video.**

`video_url` (line 318) is derived from `preview_url`, which for GCS videos is the `…/thumbs/foo.gif` URL. So when the thumb 404s — the entire reason this `<object>` fallback exists — the inner `<video src='…/thumbs/foo.gif#t=0.001'>` tries to play the same broken GIF as a video.

In master this used `output_url` (the original mp4) for the inner `<video src>`:

```python
output_url = furl(output_url).add(fragment_args={"t": "0.001"}).url
…
<video src={output_url!r} …>
```

Fix: derive `video_url` from the original `url`, never `preview_url`:

```python
video_url = furl(url).add(fragment_args={"t": "0.001"}).url
```

This also keeps the `preview_url == url` branch correct.

widgets/saved_workflow.py
```diff
@@ -309,21 +309,55 @@ def render_workflow_media(output_url: str, published_run: PublishedRun):
...
+        render_media_component(output_url)
+
+
+def render_media_component(url: str):
+    preview_url, is_video = meta_preview_url(url)
+    if is_video:
+        video_url = furl(preview_url).add(fragment_args={"t": "0.001"}).url
+        if preview_url == url:
+            gui.html(
+                f'<video src={video_url!r} class="gui-video w-100" autoplay playsInline loop muted controls>'
```
devxpy, line 321: **Missing `</video>` close tag (both branches).**

The simple branch here (line 321) and the `<object>`-wrapped branch (line 327) both emit a `<video>` with no closing tag. Browsers tolerate it, but inside `<object>` the parser keeps `<video>` open until it hits `</object>`, which can produce odd layout/parsing in edge cases. CodeRabbit flagged this on an earlier revision — worth fixing now:

```diff
- <video src={video_url!r} class="gui-video w-100" autoplay playsInline loop muted controls>
+ <video src={video_url!r} class="gui-video w-100" autoplay playsInline loop muted controls></video>
```

Same fix in the `<object>` branch around line 327.

widgets/saved_workflow.py
```diff
@@ -309,21 +309,55 @@ def render_workflow_media(output_url: str, published_run: PublishedRun):
...
+    if the thumbnail doesn't exist.
+
+    Uses <object data=thumb><img src=original/></object>. The browser tries the thumb
+    first; on 404 it renders the inner <img>. This avoids a server-side GCS existence
+    check in the page-load hot path.
+    """
+    if thumb_url == fallback_url:
+        return gui.image(src=fallback_url, style=style, className=className)
+
+    with gui.tag("object", data=thumb_url, style=style, className=className):
```
devxpy, line 362: **Click-to-open the full photo no longer works on the workflow avatar.**

`gui.image` attaches an `onClick={() => window.open(args.src)}` to the `` (see `gooey-gui/app/renderer.tsx` `case "img"`). Here that handler is only on the inner fallback ``. When `<object>` successfully renders the thumb, clicks land on the `<object>` content and don't bubble to the inner ``, so the avatar stops being clickable.

In master, `gui.image(src=pr.photo_url, …)` made the avatar clickable to open the full photo. Was that intended to be dropped, or should we wrap the `<object>` in a click handler / anchor pointing at `fallback_url`?

daras_ai_v2/base.py
```diff
@@ -2372,7 +2383,12 @@ def _render_run_preview(self, saved_run: SavedRun):
...
+                    else:
+                        self.render_run_preview_output(state)
```
devxpy, line 2391: **Heads up: this silently bypasses `render_run_preview_output` whenever `preview_image(state)` returns any URL.**

That strips richer per-recipe previews on the run preview cards. From a quick scan:

- `recipes/VideoGenPage.py` `render_run_preview_output` — multiple model outputs with prompts and download buttons → now only the first video.
- `recipes/QRCodeGenerator.py` — image + QR data caption + shortened URL → now only the bare image.
- `recipes/ObjectInpainting.py` — input + prompt + output → now only the first output image.
- `recipes/LipsyncTTS.py` — captioned video w/ download button → now bare video.
- `recipes/VideoBots.py` — input prompt + multimedia outputs → now only the first media asset.

If this is intentional (the commit message `_render_run_preview: add generic output media render` suggests it is), it'd be worth:

1. calling it out explicitly in the PR description, and
2. confirming with owners of the affected recipes — or letting recipes opt out by overriding `preview_image` to return `None` when they have a custom renderer.

devxpy:
https://github.com/GooeyAI/gooey-server/pull/962

PR #953, opened 2026-04-17
feat: bubble up insufficient credits error from function calls

nikochiko:
### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

### DIFF
--- a/functions/workflow_tools.py
+++ b/functions/workflow_tools.py
@@ -7,6 +7,7 @@ import typing
 from django.utils.text import slugify
 
 from app_users.models import AppUser
+from daras_ai_v2.exceptions import UserError
 from functions.gooey_builder_tools import UpdateGuiStateLLMTool
 from functions.inbuilt_tools import (
     FeedbackCollectionLLMTool,
@@ -132,9 +133,14 @@ class WorkflowLLMTool(BaseLLMTool):
         if self.trigger == FunctionTrigger.post:
             return
         fn_sr.wait_for_celery_result(result)
-        # if failed, raise error
+        # if failed, bubble up error so the parent run's standard error pipeline renders it
         if fn_sr.error_msg:
-            raise RuntimeError(fn_sr.error_msg)
+            raise CalledFunctionError(
+                message=fn_sr.error_msg,
+                status_code=fn_sr.error_code,
+                error_params=fn_sr.error_params,
+                error_type=fn_sr.error_type,
+            )
 
         if fn_sr.workflow != Workflow.FUNCTIONS:
             page_cls = Workflow(fn_sr.workflow).page_cls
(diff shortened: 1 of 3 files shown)


functions/recipe_functions.py
```diff
@@ -196,6 +197,8 @@ def call(self, **kwargs):
         fn_sr.wait_for_celery_result(result)
         # if failed, raise error
         if fn_sr.error_msg:
+            if fn_sr.error_type == InsufficientCredits.__qualname__:
+                raise InsufficientCredits(self.current_user, self.saved_run)
```
devxpy: instead of special casing here, bubble up error_msg, error_type & error_params to _render_failed_output(), _render_custom_error() for using standard error rendering pipeline

nikochiko: added.
i had thought of this but because the `__init__` params are different for the different subclasses of `UserError`, we can't dynamically replicate the error objects without inspecting annotations / switch-case. Added `error_type` on `UserError` as a workaround to this.

a better solution though would be to standardise the `__init__` params (to accept `error_params`) and let the `render` method accept `request`, etc. when they are needed.

with that, we could write

```python
if fn_sr.error_msg:
  exc_cls = getattr(exceptions, fn_sr.error_type, None)
  if exc_cls is not None:
    raise exc_cls(fn_sr.error_params)
  raise RuntimeError(fn_sr.error_msg)
```

functions/recipe_functions.py
```diff
@@ -194,8 +192,16 @@ def call(self, **kwargs):
...
         if fn_sr.error_msg:
+            exc_cls = getattr(exceptions, fn_sr.error_type, None)
+            if exc_cls and issubclass(exc_cls, UserError):
+                raise UserError(
+                    fn_sr.error_msg,
+                    status_code=fn_sr.error_code,
+                    error_params=fn_sr.error_params,
+                    error_type=fn_sr.error_type,
+                )
             raise RuntimeError(fn_sr.error_msg)
```
devxpy: I think the RuntimeError here is not particularly useful anyways so better to chuck it, lets do this:

```
class CalledFunctionError(UserError):
      ...

if fn_sr.error_msg:
   raise CalledFunctionError(
        fn_sr.error_msg,
        status_code=fn_sr.error_code,
        error_params=fn_sr.error_params,
        error_type=fn_sr.error_type,
    )
```

(and remove error_type from base UserError)

functions/workflow_tools.py
```diff
@@ -132,9 +133,14 @@ def call(self, **kwargs):
...
-        # if failed, raise error
+        # if failed, bubble up error so the parent run's standard error pipeline renders it
         if fn_sr.error_msg:
-            raise RuntimeError(fn_sr.error_msg)
+            raise CalledFunctionError(
+                message=fn_sr.error_msg,
+                status_code=fn_sr.error_code,
+                error_params=fn_sr.error_params,
+                error_type=fn_sr.error_type,
+            )
```
nikochiko, line 143: this raises ALL errors from function calls as a `UserError` (because `CalledFunctionError` subclasses from `UserError`), including other errors like API failures / gooey bugs. 
i think we want to either:
- not subclass `CalledFunctionError` from `UserError`
- check `error_type` is a subclass of `UserError` before raising it as `CalledFunctionError`

devxpy: Yes, that is good. With current behavior, other errors are caught twice in sentry. making them UserError catches them only once. We could also possibly include a link in the err message to the tool call

devxpy: you can also special-case CalledFunctionError in runner_task, or define a new parent class of both UserError and CalledFunctionError - but both effectively do the same job

PR #969, opened 2026-04-30
feat: when filtering by workspace default sort should be last_updated

nikochiko:
### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

### DIFF
--- a/widgets/workflow_search.py
+++ b/widgets/workflow_search.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import re
 import typing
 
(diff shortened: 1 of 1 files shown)


widgets/workflow_search.py
```diff
@@ -154,11 +159,12 @@ def render_search_filters(
...
                     sort_options = {
-                        opt.name if opt != SortOptions.get() else None: (
-                            opt.html_icon_label()
-                        )
-                        for opt in SortOptions
+                        option.name: option.html_icon_label() for option in SortOptions
                     }
+                    default_sort_option = SortOptions.get(search_filters=search_filters)
+                    if search_filters.sort != default_sort_option.name:
+                        sort_options[""] = sort_options.pop(default_sort_option.name)
```
nikochiko, line 166: explanation:
in the case that we have a `sort=featured` / `sort=last_updated` param, we should persist that after a change in the workspace filter because a user has explicitly set the sort criteria to that. in the case where sort parameter is absent, or set to some other value, we should stick with the default

devxpy (approved):

PR #971, opened 2026-05-04
feat: add a7, a30, r7, r30 columns for messages export in CSV

nikochiko:
### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

### DIFF
--- a/bots/models/convo_msg.py
+++ b/bots/models/convo_msg.py
@@ -9,7 +9,7 @@ import pytz
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db import models
-from django.db.models import OuterRef, Q, QuerySet, Subquery
+from django.db.models import Count, Max, Min, OuterRef, Q, QuerySet, Subquery
 from django.utils.text import Truncator
 
 from bots.custom_fields import CustomURLField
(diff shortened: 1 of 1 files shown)


devxpy (commented):
<>

devxpy (commented):
This looks like it repeats the same conversation-level metrics logic that already exists in `ConversationQuerySet.to_df()` for A7/A30/R1/R7/R30 and thumbs up/down counts.

Could we extract this into a shared helper/queryset annotation and use it from both the conversation export and message export paths? That would avoid keeping two implementations of the same metric definitions that can drift over time.

devxpy:
I also verified the CSV path with a small Django script: I monkeypatched `MessageQuerySet.to_json()` to return a row containing `A7`, `A30`, `R1`, `R7`, `R30`, `thumbs_up`, and `thumbs_down`, then called the real `Message.objects.none().to_df()` implementation.

The resulting DataFrame columns were only:

`Sent`, `Name`, `User Message (EN)`, `Assistant Message (EN)`, `User Message (Local)`, `Assistant Message (Local)`, `Analysis Result`, `Feedback`, `Run Time`, `Credits Used`, `Run URL`, `Input Images`, `Input Audio`, `User Message ID`, `Conversation ID`, `Integration Name`

So the new fields are currently computed by `to_json()` but dropped before `df.to_csv()` because `MessageQuerySet.to_df()` builds a fixed row dict and does not include them.

nikochiko:
thanks! i've made the changes

devxpy (approved):

PR #972, opened 2026-05-05
fix: raw_tts_text cleanup for thinking traces

milovate:
### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

### DIFF
--- a/recipes/VideoBots.py
+++ b/recipes/VideoBots.py
@@ -815,15 +815,16 @@ Translation Glossary for LLM Language (English) -> User Langauge
                 model=request.translation_model,
             )
             # save translated response for tts
-            response.raw_tts_text = [
+            tts_source = [
                 "".join(snippet for snippet, _ in parse_refs(text, response.references))
                 for text in output_text
             ]
+            response.raw_tts_text = tts_source
+        else:
+            tts_source = output_text
 
         # remove html tags from the output text for tts
-        raw_tts_text = [
-            parse_bot_html(text)[1] for text in response.raw_tts_text or output_text
-        ]
+        raw_tts_text = [parse_bot_html(text)[1].strip() for text in tts_source]
         if raw_tts_text != output_text:
             response.raw_tts_text = raw_tts_text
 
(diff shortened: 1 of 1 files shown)


milovate:
@codex review

devxpy:
The root cause here is just that `response.raw_tts_text or output_text` reads back a value `output_translation_step` itself wrote on a previous streaming chunk. During streaming, an early chunk like `<think>...incomplete...` parses to `''` under pyquery (everything is swallowed by the unclosed `<think>` element), so `response.raw_tts_text` gets set to `['']` and then `[''] or output_text` is truthy on every later chunk — that's why it stays empty even after the closing tags arrive.

Once you stop reading back `response.raw_tts_text` as the source within this function, the bug goes away — no `zip_longest`, no double-`parse_bot_html`, no extra `None` reset. Just use a local source:

```diff
     def output_translation_step(self, request, response, output_text):
         from daras_ai_v2.bots import parse_bot_html

         # translate response text
         if should_translate_lang(request.user_language):
             yield f"Translating response to {request.user_language}..."
             output_text = run_translate(
                 texts=output_text,
                 source_language="en",
                 target_language=request.user_language,
                 glossary_url=request.output_glossary_document,
                 model=request.translation_model,
             )
             # save translated response for tts
-            response.raw_tts_text = [
+            tts_source = [
                 "".join(snippet for snippet, _ in parse_refs(text, response.references))
                 for text in output_text
             ]
+            response.raw_tts_text = tts_source
+        else:
+            # avoid carrying over a stale raw_tts_text from an earlier streaming
+            # chunk (e.g. one that contained an unclosed <think> tag and parsed
+            # to '' under pyquery), which would otherwise stick because
+            # `[''] or output_text` is truthy.
+            tts_source = output_text

         # remove html tags from the output text for tts
-        raw_tts_text = [
-            parse_bot_html(text)[1] for text in response.raw_tts_text or output_text
-        ]
+        raw_tts_text = [parse_bot_html(text)[1].strip() for text in tts_source]
         if raw_tts_text != output_text:
             response.raw_tts_text = raw_tts_text

         return output_text
```

Why this is better than the current diff:

- Smaller, surgical change — fixes the actual bug (stale-read via `or`) instead of papering over it with a fallback re-parse.
- No `zip_longest` + second `parse_bot_html(fallback_text)` pass. That fallback only kicks in when the translated/source text strips to empty, which in practice means the *original* output is purely a `<think>` block + buttons. In that case TTS-ing the raw output (which still contains the unclosed/think text) is arguably worse than empty.
- Doesn't need `response.raw_tts_text = None` reset, because we no longer pollute it from a partial chunk in the first place.
- Keeps the `raw_tts_text != output_text` invariant intact (only set when actually different).
- `.strip()` on the parsed result is a good addition — kept it.

Separately, the `get_raw_price` change (`state.get("raw_tts_text") or state.get("raw_output_text", [])`) is a nice independent fix and worth keeping.

devxpy (approved):

PR #992, opened 2026-06-01
feat: new /home landing page for logged-in users

anish-work:
Summary
Adds a new /home page that replaces "Saved" as the primary logged-in entry point in the global header. The page surfaces the user's recent runs and saved workflows alongside admin-curated content (workflow tabs, industry tiles, news feed). Workspace-aware: in a team workspace it shows a workspace header (logo, name, description, settings link for editors) and scopes recent workflows to the current user's runs.

Changes
New cms Django app with admin-curated models: WorkflowTab, WorkflowCard, IndustryTile, NewsItem — all priority-ordered, hideable via priority=0.
widgets/home.py loads data (recent/saved workflows, tabs, tiles, news) and hands typed props to the React component.
React HomePage component tree under [gooey-gui/app/components/HomePage/](vscode-webview://0qlc7pusej3c5umfcj3i60a5trv6tm3o7g2vouaudlfg0mvo0705/gooey-gui/app/components/HomePage/) — index.tsx, WorkspaceHeader.tsx, IndustryBrowser.tsx, NewsFeed.tsx, workflows.tsx (history / saved / picker card variants + PreviewContent helper), plus HomePage.css.
Routing: new /home/ route in [routers/root.py](vscode-webview://0qlc7pusej3c5umfcj3i60a5trv6tm3o7g2vouaudlfg0mvo0705/routers/root.py); global header now links logged-in users to Home (Saved is dropped from the header).
Model tweaks: Tag.description field (migration [bots/migrations/0119_tag_description.py](vscode-webview://0qlc7pusej3c5umfcj3i60a5trv6tm3o7g2vouaudlfg0mvo0705/bots/migrations/0119_tag_description.py)); PublishedRun.get_share_badge_data() returns structured {iconHtml, label} for the frontend; Workflow.get_metadata() non-creating accessor.

### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

---
  - https://app.asana.com/0/0/1214767606080310
  - https://app.asana.com/0/0/1215108872315142


widgets/demo_button.py
```diff
@@ -50,6 +57,37 @@ def get_demo_bots(pr: PublishedRun):
     )
 
 
+def demo_platform_ids_by_pr_id(pr_ids: list[int]) -> dict[int, list[int]]:
```
devxpy: **Dead code?** These four new symbols (`PlatformPillData`, `demo_platform_ids_by_pr_id`, `platform_pill_data`, `platform_pills_for_ids`) don't appear to be referenced anywhere in the repo — a clone-wide grep finds only the definitions here. They aren't imported by `widgets/home.py` or anything else. If they're leftover scaffolding from an earlier iteration of the home cards, can we drop them (or wire them up if still intended)?

gooey-gui/app/components/HomePage/workflows.tsx
```diff
@@ -0,0 +1,323 @@
...
+      )}
+    </a>
+  );
+}
+
+export function WorkflowPickerCard({ card }: { card: CardData }) {
+  const hasAuthor = !!(card.authorPhotoUrl || card.authorName);
+  const fallbackLetter = (card.title.trim().charAt(0) || "?").toUpperCase();
+
+  const hasFooter =
```
devxpy: **Picker footer fields are never populated.** `WorkflowPickerCard` renders `updatedAt`, `runCount`, and `accessBadge` in the footer, but in `widgets/home.py::_workflow_to_json` the `picker` profile sets none of them:

- `updatedAt` is guarded by `if updated_at and profile != "picker"`
- `runCount` / `accessBadge` are only set under `if profile == "saved"`

So for picker cards `hasFooter` can only be true via `hasAuthor`, and the clock / runs / badge markup is effectively dead. Either populate these fields for picker cards in the backend, or remove the markup here to avoid the misleading contract.

widgets/home.py
```diff
@@ -0,0 +1,463 @@
...
+            profile="saved",
+            pr=pr,
+            author=pr.last_edited_by,
+            current_user=user,
+        )
+        for pr in prs
+    ]
+
+
+def _load_workflow_tabs() -> list[WorkflowTabData]:
```
devxpy: **Unbounded cards per tab → N+1 on metadata.** Unlike recent/saved (capped at 3–4), there's no limit on cards per tab here. For each card, `_workflow_to_json` calls `workflow.get_metadata()`, which issues a fresh `WorkflowMetadata.objects.filter(...).first()` query with no caching — and re-queries for every card even when they share the same workflow. A tab with N cards = N metadata queries. Consider capping cards per tab and/or memoizing `get_metadata()` by workflow for the duration of the render.

routers/root.py
```diff
@@ -321,6 +321,18 @@ def explore_page(
     }
 
 
+@gui.route(app, "/home/")
+def home_page(request: Request):
```
devxpy, line 337: **Anonymous `/home/` shows a near-blank page.** This route is public, but the header only links logged-in users here. An anonymous visitor to `/home/` gets no greeting, recent, or saved sections (workspace is `None`) — just tabs/tiles/news. Should we redirect anonymous users to login (or `/explore/`) instead of rendering a partial page?

gooey-gui/app/components/HomePage/workflows.tsx
```diff
@@ -0,0 +1,323 @@
...
+
+export function HistoryWorkflowCard({ card }: { card: CardData }) {
+  const hasAuthor = !!(card.authorPhotoUrl || card.authorName);
+  return (
+    <a
+      href={card.href}
+      className="d-flex flex-column h-100 border rounded-4 overflow-hidden text-decoration-none text-body border-hover"
+    >
+      <div className="position-relative recent-card-preview bg-light p-0">
+        {(card.workflowLabel || card.workflowEmoji) && (
```
devxpy: **Empty badge when label but no emoji.** The badge wrapper renders on `(card.workflowLabel || card.workflowEmoji)`, but the inner span only renders `card.workflowEmoji`. If a workflow has a `workflowLabel` but no emoji, an empty badge pill is shown. Also `workflowLabel` is set in the payload but never displayed anywhere. Either render the label or drop it from the payload and tighten this condition to `card.workflowEmoji`.

cms/models.py
```diff
@@ -0,0 +1,90 @@
...
+
+class NewsItem(models.Model):
+    headline = models.CharField(max_length=200)
+    tag = models.CharField(
+        max_length=64,
+        help_text='Freeform label, e.g. "Event", "Product", "Case Study".',
+    )
+    photo_url = CustomURLField(default="", blank=True)
+    publish_date = models.DateField()
+    url = models.CharField(max_length=500, help_text="Absolute URL the card links to.")
```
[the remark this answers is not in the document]
devxpy: Use CustomURLField

cms/models.py
```diff
@@ -0,0 +1,90 @@
...
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        ordering = ["-priority", "id"]
+
+    def __str__(self):
+        return f"{self.workflow_tab.title} — {self.recipe}"
+
+
+class IndustryTile(models.Model):
```
devxpy: **Is `IndustryTile` needed, or is this just `Tag`?** This model is effectively `FK(Tag) + priority`, and `_load_industry_tiles` pulls every display field (`name`, `icon`, `description`) off `tile.tag` — the tile only contributes an ordering value that duplicates `Tag.featured_priority`. `Tag` already has everything required:

- `category` with a dedicated `TagCategory.industry` choice to select industry tags
- `featured_priority` for ordering + hiding (`=0`)
- `icon`, `description` (the latter added in this PR via migration `0119`)

We could drop the model + migration + admin + join and drive the section directly, e.g.:

```python
Tag.objects.filter(category=TagCategory.industry, featured_priority__gte=1)
   .annotate(workflow_count=Count(...))
   .order_by("-featured_priority", "name")
```

The only reason to keep `IndustryTile` is if product wants the home-page industry ordering to be independent of `Tag.featured_priority` (which also drives tag ordering elsewhere). If that's the goal, a dedicated `industry_priority` field on `Tag` is lighter than a whole join model. Otherwise, suggest collapsing this into `Tag` + `category=industry`.

cms/models.py
```diff
@@ -0,0 +1,90 @@
...
+    icon = models.TextField(blank=True, default="")
+    priority = models.IntegerField(
+        default=1,
+        help_text="Higher priority tabs are shown first. If 0, then the tab is not shown at all.",
+    )
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        ordering = ["-priority", "id"]
```
devxpy, line 18: **Prefer dropping `Meta.ordering` entirely and ordering at the query level instead.** All four cms models define a default `ordering` (`["-priority", "id"]` here, and `["-publish_date", "-id"]` on `NewsItem` at L86), but `widgets/home.py` already overrides them with explicit `.order_by(...)` anyway — so the model-level ordering is both redundant on the hot path and inconsistent (asc `id` here vs `-id` on `NewsItem`).

Implicit default ordering also has a habit of silently attaching itself to unrelated queries (counts, `.exists()`, joins) and hurting performance. Suggest removing `Meta.ordering` from `WorkflowTab`, `WorkflowCard`, `IndustryTile`, and `NewsItem`, and making each call site order explicitly with a deterministic tiebreak.

widgets/home.py
```diff
@@ -0,0 +1,463 @@
...
+            "tag": item.tag,
+            "photoUrl": item.photo_url or None,
+            "age": _format_news_age(item.publish_date),
+            "href": item.url,
+        }
+        for item in qs
+    ]
+
+
+def _format_news_age(publish_date: date) -> str:
```
devxpy: **Reuse the existing relative-time helper instead of reimplementing it.** This duplicates `get_relative_time` in `daras_ai_v2/utils.py` — which this file *already imports* and uses for `updatedAt`. Both even share the same `d`/`mo`/`y` unit vocabulary.

Two caveats so it isn't a blind swap:

- **Format differs slightly:** `_format_news_age` has a `w` (weeks) bucket and no suffix; `get_relative_time` has no weeks bucket and appends `" ago"` (so "2w" vs "14d ago").
- **`date` vs tz-aware `datetime`:** `get_relative_time` does `timezone.now() - timestamp` and expects a tz-aware `datetime`, while `publish_date` is a `DateField` — would need conversion first.

If the compact, suffix-less "2w" format is a deliberate design choice for news cards, that's fine — but a one-line comment saying so would prevent the "don't we already have this?" question. (`django.contrib.humanize.naturaltime`, used elsewhere in the repo, is another option if humanized text is acceptable.)

cms/models.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class NewsItem(models.Model):
+    headline = models.CharField(max_length=200)
+    tag = models.CharField(
+        max_length=64,
+        help_text='Freeform label, e.g. "Event", "Product", "Case Study".',
+    )
+    photo_url = CustomURLField(default="", blank=True)
+    publish_date = models.DateField()
```
devxpy: Why `DateField` and not `DateTimeField`? A datetime would:

- let you reuse `get_relative_time(item.publish_date)` directly (it expects a tz-aware `datetime`), removing the need for the custom `_format_news_age` + `date`→`datetime` conversion noted above
- support scheduling a publish at a specific time, and a precise `publish_date__lte=now()` gate rather than whole-day granularity

`date_hierarchy` / `list_display` in the admin work fine with `DateTimeField` too. Worth switching here (and in migration `0001`) unless there's a reason it must be date-only.

widgets/home.py
```diff
@@ -0,0 +1,463 @@
...
+        "userMessage": _preview_text(user_message, CHAT_PREVIEW_MAXLEN),
+        "botMessage": _preview_text(bot_message, CHAT_PREVIEW_MAXLEN),
+    }
+
+
+def _media_preview(
+    *,
+    output_url: str,
+    state: dict | None = None,
+    page_cls=None,
```
devxpy: **Missing type annotation on `page_cls`.** It should be `type[BasePage] | None` (`Workflow.page_cls` returns `typing.Type[BasePage]`). To keep this off the module import path — the PR checklist explicitly tracks import time, and `daras_ai_v2.base` is heavy — add it under a `TYPE_CHECKING` guard

```python
from __future__ import annotations


if typing.TYPE_CHECKING:
    from daras_ai_v2.base import BasePage

def _media_preview(*, output_url: str, state: dict | None = None,
                   page_cls: type[BasePage] | None = None,
                   caption: str | None = None) -> dict:
```

Same `page_cls` param is untyped in `_sr_preview` / `_pr_preview` — worth annotating those for consistency too.

widgets/home.py
```diff
@@ -0,0 +1,463 @@
...
+    return data
+
+
+def _sr_preview(
+    *,
+    workflow: Workflow,
+    sr: SavedRun,
+    pr: PublishedRun | None,
+    metadata: WorkflowMetadata | None,
+    include_chat: bool,
```
devxpy: `include_chat` is always `True` — the only call site (`_workflow_to_json`, L262) passes `include_chat=True` and there are no others. So the parameter is dead and `if include_chat and workflow == Workflow.VIDEO_BOTS:` reduces to `if workflow == Workflow.VIDEO_BOTS:`. Suggest dropping the param and simplifying the condition, unless a caller that passes `False` is coming in a follow-up.

widgets/home.py
```diff
@@ -0,0 +1,463 @@
...
+CHAT_PREVIEW_MAXLEN = 130
+MEDIA_CAPTION_MAXLEN = 60
+
+
+class AccessBadgeData(TypedDict):
+    iconHtml: str
+    label: str
+
+
+class WorkflowCardData(TypedDict, total=False):
```
devxpy: **Consider Pydantic `BaseModel` instead of `TypedDict` here.** This file already has a live instance of the failure mode: L306 does `data["changeNotes"] = change_notes`, but `changeNotes` isn't a declared key on `WorkflowCardData` (the React `CardData` type *does* have `changeNotes?: string`). A type checker flags `TypedDict "WorkflowCardData" has no key "changeNotes"`, but at runtime it's just a dict so the typo/missing-key sails through — and `total=False` means even missing *required* fields aren't caught.

Pydantic models are already the idiomatic choice in this repo (recipe `RequestModel`/`ResponseModel`), and would give construction-time validation:

```python
class WorkflowCardData(BaseModel):
    model_config = ConfigDict(extra="forbid")
    title: str
    href: str
    change_notes: str | None = None  # alias="changeNotes"
    ...
```


Minimum fix regardless of approach: add `changeNotes` to the model/TypedDict. Also note `_load_industry_tiles` (L408) and `_load_news_items` (L438) return bare `list[dict]` with no typing at all — they'd benefit from the same modeling.

widgets/home.py
```diff
@@ -0,0 +1,463 @@
...
+        for tab in qs
+    ]
+
+
+def _workflow_to_json(
+    *,
+    profile: WorkflowCardProfile,
+    pr: PublishedRun | None = None,
+    sr: SavedRun | None = None,
+    author: AppUser | Workspace | None = None,
```
devxpy: Agreed, the `author: AppUser | Workspace | None` union is a smell — it forces runtime `isinstance` dispatch (L282-291) and the two types don't even share a shape: `Workspace.display_name()` is a method + `get_photo()`, whereas `AppUser.display_name` is an attribute + `photo_url`. That shape mismatch is why a `Protocol` won't cleanly unify them.

Since every call site already knows the concrete type — `_load_recent_workflows`/`_load_saved_workflows` pass an `AppUser`, `_load_workflow_tabs` passes a `Workspace` — the cleanest fix is to normalize to author display fields *before* this function and pass a single small value object:

```python
class AuthorData(BaseModel):
    name: str
    photo_url: str | None = None

def author_from_user(u: AppUser, current_user: AppUser | None) -> AuthorData: ...
def author_from_workspace(w: Workspace) -> AuthorData: ...
```

Then `_workflow_to_json` takes `author: AuthorData | None` and just reads `.name` / `.photo_url` — no union, no `isinstance`. (At minimum, extract the L282-291 block into a typed `_resolve_author(author, current_user) -> AuthorData | None` helper.)

More broadly this function is carrying a lot: a `pr | sr` union, a `profile` discriminator, and this `author` union all in one body. Worth considering a split per profile, but the `author` normalization is the highest-value piece to pull out.

widgets/home.py
```diff
@@ -0,0 +1,463 @@
...
+        title=META_TITLE,
+        description=META_DESCRIPTION,
+    )
+
+
+def _get_greeting(user: AppUser) -> str | None:
+    return user.first_name(fallback="") or None
+
+
+def _load_recent_workflows(
```
devxpy, line 195: **This will be slow on prod — measured 2.39s on the heaviest workspace, and it runs on every home-page load.** I profiled it against prod (`EXPLAIN ANALYZE`, live DB).

The `DISTINCT ON (parent_version__published_run_id)` forces Postgres to read **all** of a workspace’s matching `SavedRun`s and disk-sort them just to emit 4 cards. `bots_savedrun` is 7.4M rows, and run history is heavily concentrated per workspace:

| workspace | runs (w/ parent PR) | distinct published runs |
|---|---:|---:|
| 5 | 1,780,569 | 1 |
| 232,791 | 231,343 | 3 |
| 3,206,451 | 69,119 | 1 |
| 3,764,345 | 53,981 | 3 |
| 3,466,675 | 52,471 | 10 |

Plan for workspace 5:

```
Limit (rows=4) actual time=2370..2383
  -> Sort (updated_at DESC)
    -> Unique  [DISTINCT ON published_run_id]
      -> Gather Merge rows=1780569
        -> Sort Key: published_run_id, updated_at DESC
              Sort Method: external merge  Disk: ~21MB
          -> Parallel Seq Scan on bots_savedrun  Filter:(workspace_id=5)  Rows Removed: 1775903
Execution Time: 2390 ms (+555ms JIT)
```

Crucially, no query reformulation fixes this: deriving “recent distinct workflows” from `SavedRun` history is inherently O(total runs), and the data above shows workspaces with millions of runs but only 1–3 distinct published runs (so a “walk -updated_at and dedup in Python” approach would scan the *entire* history and still never reach 4 distinct).

**Proper fix: denormalize last-used.** Maintain a small `(workspace, published_run, last_used_at)` row, upserted whenever a run is saved (and a backfill migration). Then recent workflows becomes:

```python
WorkspaceRecentWorkflow.objects.filter(workspace=workspace)
    .order_by("-last_used_at")[:RECENT_WORKFLOW_LIST_LIMIT]
```

That reads 1–240 rows per workspace instead of up to 1.78M — O(distinct), exact, and index-friendly (`Index(workspace, -last_used_at)`; add `uid` for the team-scoped path). Happy to share the EXPLAIN output if useful.

widgets/home.py
```diff
@@ -0,0 +1,463 @@
...
+        else None
+    )
+    workspace_header = _get_workspace_header(request.user, workspace)
+    gui.component(
+        "HomePage",
+        greeting=(
+            _get_greeting(request.user)
+            if not is_anonymous and workspace_header is None
+            else None
+        ),
```
devxpy: Nit: this nested ternary inside the `gui.component(...)` kwargs is hard to parse (the `not is_anonymous and workspace_header is None` condition especially). Lift it into a plain `if/else` above the call:

```python
if not is_anonymous and workspace_header is None:
    greeting = _get_greeting(request.user)
else:
    greeting = None

gui.component(
    "HomePage",
    greeting=greeting,
    workspaceHeader=workspace_header,
    ...
)
```

Reads more clearly and keeps the `gui.component(...)` call to simple keyword args.

gooey-gui/app/styles/custom.css
```diff
@@ -911,3 +929,20 @@ button {
 .ag-horizontal-left-spacer, .ag-body-viewport {
   background-color: #f7f7f7;
 }
+
+/* Bootstrap 5.3 object-fit utilities (we're on 5.2.3) */
```
devxpy: Why backport these instead of bumping Bootstrap? The version is already inconsistent across the repo, and it’s a CDN `<link>` (no npm/build dependency), so upgrading is a one-liner:

- `gooey-gui/app/app.tsx` loads `bootstrap@5.2.3` (this is the app HomePage renders in)
- `templates/base.html` and `auto_payment_failed_email.html` already load `bootstrap@5.3.0`

Since 5.3 is already trusted elsewhere, changing the `app.tsx` CDN URL `5.2.3` → `5.3.x` gives native `object-fit-*` utilities for free and lets this whole block go away (also aligning the app with `base.html`).

Caveat: 5.3 added color modes (dark mode) and some `--bs-*` variable changes, so bumping the *global* stylesheet could shift visuals app-wide, not just on HomePage — worth a quick visual smoke test. If you’d rather not take that risk in this PR, keeping the scoped backport is defensible, but then please drop the “we’re on 5.2.3” framing and note that `app.tsx` vs `base.html` are intentionally on different versions.

widgets/home.py
devxpy: consider shipping without this section if needed

gooey-gui/app/components/HomePage/types.ts
```diff
@@ -0,0 +1,73 @@
+export type ChatPreview = {
```
devxpy, line 1: **Potential exploration (not for this PR): generate these TS types from the Python props instead of hand-maintaining them.** This file is a hand-written mirror of the Python payload, which is exactly how the `changeNotes` drift (declared in TS, missing from the Python `WorkflowCardData`) slipped in. Worth considering a codegen pipeline so the two can’t diverge.

The repo is already most of the way there: Pydantic + `model_json_schema()` is used everywhere (`recipes/*`, `daras_ai_v2/base.py`, ref handling in `daras_ai_v2/openapi_tricks.py`), and FastAPI already serves `/openapi.json`. The only missing piece is the TS generation step — there’s no `openapi-typescript` / `json-schema-to-typescript` in `gooey-gui/package.json` today.

Sketch:

1. Define each component’s props as a Pydantic `BaseModel` (the `WorkflowCardData` → `BaseModel` change suggested elsewhere).
2. Codegen script: collect the prop models → `model_json_schema(ref_template=...)` (or wrap into an OpenAPI `components.schemas` doc).
3. Run `openapi-typescript` (OpenAPI) or `json-schema-to-typescript` (raw JSON Schema) → emit a generated `.d.ts`.
4. Components import the generated types instead of this file.
5. **CI drift check**: regenerate + `git diff --exit-code` so a Python-side change not reflected in TS fails CI. That’s the part that actually prevents the drift.

**Casing:** keep everything camelCase on both sides (it already is, e.g. `workflowLabel`, `authorPhotoUrl`, `iconHtml`). That makes the Pydantic conversion a verbatim swap (no `alias_generator`/`by_alias`), keeps the wire JSON byte-identical, and needs no React/`types.ts` edits — the only cost is a scoped `N815` (mixed-case attr) lint ignore on the props modules. The discriminated `CardPreview` union maps cleanly via Pydantic `Field(discriminator="type")` → `oneOf`+`discriminator` → a TS discriminated union.

Repo-wide infra (touches the ~dozen components in `app/components/index.ts`), so a separate task — but `HomePage` would be a good proof-of-concept.

devxpy: @milovate

anish-work:
### Addressed updates in this branch based on review feedback:
- Switched `NewsItem.url` to `CustomURLField` and kept migration aligned (`f495525c`).
- Moved `NewsItem.publish_date` to `DateTimeField` and reused relative-time handling in home feed (`b07b4791`).
- Added request-scoped metadata memoization in `widgets/home.py` to remove repeated `workflow.get_metadata()` lookups (`a077599a`).
- Reworked home payload typing from `TypedDict` to Pydantic models (including `changeNotes`) so backend/frontend props stay consistent (`6dee6c60`).
- Added empty-tabs guard in `WorkflowPicker` (`401c7756`).
- Simplified conditional flow in `home.render()` to explicit `if/else` style (`0fb88657`).
- Refactored card builders per profile for clearer, safer payload construction (`abe1d621`).


### Product decisions confirmed with @SeanBlagsvedt  and reflected:
- We are intentionally keeping **Industry Tiles as a separate CMS model** (`IndustryTile`) and not collapsing into `Tag`.
- We are intentionally keeping **`/home` behavior for anonymous users as-is** (no redirect change in this PR).


### Follow-up:
- Recent-workflow performance is being tracked separately in #993.

gooey-gui/app/components/HomePage/types.ts
milovate: would incorporate this in  https://github.com/GooeyAI/gooey-server/pull/987

widgets/home.py
```diff
@@ -0,0 +1,565 @@
...
+            href=_industry_tile_href(tag),
+        )
+        for tag in qs
+    ]
+
+
+def _industry_tile_href(tag: Tag) -> str:
+    if tag.landing_page:
+        return tag.landing_page
+    return str(furl("/explore/", query_params={"search": tag.name}))
```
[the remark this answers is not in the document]
anish-work: The priorities have changed, we have dropped the IndustryTile model entirely

bots/migrations/0122_tag_featured_tag_hide_tag_landing_page_and_more.py
```diff
@@ -0,0 +1,47 @@
...
+            ),
+        ),
+        migrations.AlterField(
+            model_name="tag",
+            name="featured_priority",
+            field=models.IntegerField(
+                default=1, help_text="Higher priority tags are shown first."
+            ),
+        ),
+    ]
```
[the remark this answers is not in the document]
anish-work: none of the tags have 0 yet, chill

anish-work:
@devxpy can you review this? I removed industry tiles and added new fields in tags, if this okay can you release this too

widgets/home.py
devxpy: ??

anish-work:
@devxpy rebased with master

widgets/home.py
anish-work: The recent-workflows table needs to be keyed by `(workspace, uid, published_run, last_saved_run, last_used_at)`, not just `(workspace, published_run, last_used_at)`, and should store `last_saved_run` + `last_used_at`.

The team home page shows "the workflows *I* recently ran." If the table is keyed only by `(workspace, published_run)`, `last_used_at` is workspace-global, so both the *selection* and *ordering* of workflows reflect whatever anyone on the team ran — it can surface workflows this user never ran (empty/<4 cards) and order them by the team's recency instead of the user's.

Adding `uid` to the table — `(workspace, uid, published_run, last_saved_run, last_used_at)` with index `(workspace, uid, -last_used_at)` — makes the read return exactly this user's recent workflows in the right order. The personal page filters by `workspace` only (same index prefix) and dedups on `published_run`.

`last_saved_run` should be stored because the card renders the user's actual run, and there's no direct FK from a published run to "this user's latest execution of it" (`PublishedRun.saved_run` points to the published/template run, not the user's). Storing it lets the card render via a single FK join.

This stays cheap: rows scale with users × distinct workflows (a handful per workspace), and reads hit the index for ~4 rows regardless of table size.

@devxpy

PR #1004, opened 2026-06-11
fix: run Composio meta tools in a per-conversation tool-router session

milovate:
COMPOSIO_* tools require a Composio tool-router session, so persist composio_session_id on Conversation and route meta-tool calls through composio.use while keeping app tools on tools.execute.

### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

### DIFF
--- a/functions/composio_tools.py
+++ b/functions/composio_tools.py
@@ -102,31 +102,6 @@ def is_composio_meta_tool(slug: str) -> bool:
     return slug.startswith("COMPOSIO_")
 
 
-def render_composio_meta_tool_scope_warning(functions: list[dict]) -> None:
-    # Meta tools share one Composio tool-router session per run, keyed to the first
-    # caller's scope.
-    from functions.base_llm_tool import get_external_tool_slug_from_url
-    from functions.models import FunctionScopes
-
-    composio_meta_tool_scopes = set()
-    for function in functions:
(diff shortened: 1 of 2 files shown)


milovate:
@codex review

milovate:
@cursor review

devxpy:
having a conversation_id means that this only works for deployments? I think the COMPOSIO_SEARCH_TOOLS will return a session_id so we can just use that in-process inside the celery task

functions/composio_tools.py
```diff
@@ -81,6 +98,50 @@ def call(self, **kwargs) -> dict:
...
+
+def render_composio_meta_tool_scope_warning(functions: list[dict]) -> None:
+    # Meta tools share one Composio tool-router session per run, keyed to the first
+    # caller's scope.
+    from functions.base_llm_tool import get_external_tool_slug_from_url
+    from functions.models import FunctionScopes
+
+    composio_meta_tool_scopes = set()
+    for function in functions:
+        composio_tool_slug = get_external_tool_slug_from_url(
```
devxpy: i think doing this in the hot path of the render would make the page slow, maybe we can just skip this warning for now and ship

devxpy (approved):

PR #1005, opened 2026-06-12
fix: migrate OpenAI realtime websocket to GA

milovate:
- OpenAI rejected Realtime websocket connections that still sent the beta negotiation header, and the old flat session/event shape could fail after the connection moved to the GA interface.

- Remove the beta header, send the session audio config in the GA shape, and keep legacy event names alongside GA names so existing streaming and Twilio paths continue to work.

https://developers.openai.com/api/docs/guides/realtime#beta-to-ga-migration

### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/language_model.py
+++ b/daras_ai_v2/language_model.py
@@ -421,7 +421,6 @@ def _run_chat_model(
                 audio_url=audio_url,
                 audio_session_extra=audio_session_extra,
                 messages=messages,
-                temperature=temperature,
                 tools=tools,
             )
         case ModelProvider.openai:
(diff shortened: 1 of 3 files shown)


milovate:
@codex review

devxpy (approved):
LGTM — the beta→GA migration matches OpenAI's official migration guide (beta header removed, `session.type: "realtime"`, nested `audio.input`/`audio.output`, `audio/pcmu` format objects, and the `response.output_*` event renames). Verified the migration is complete across the codebase (no lingering old event names or beta header), and CI is green.

One follow-up (non-blocking): we should stop accepting `temperature` in `run_openai_audio`. GA removed `temperature` as a model parameter, so it's correctly gone from the session payload — but `run_openai_audio` still clamps it to the old beta range `(0.6, 1.2)` and threads it into `init_ws_session`, which no longer uses it. It's now dead plumbing. Recommend dropping the clamp and the `temperature` param from `init_ws_session` (and the `temperature=` kwarg at the call site in `language_model.py`) so it's clear the value is intentionally not sent.

PR #1006, opened 2026-06-12
fix: OpenAI TTS voice exclude unsupported voices based on the selecte…

milovate:
…d model.

### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

---
  - https://app.asana.com/0/0/1210195680779266

### DIFF
--- a/daras_ai_v2/text_to_speech_settings_widgets.py
+++ b/daras_ai_v2/text_to_speech_settings_widgets.py
@@ -75,6 +75,17 @@ class OpenAI_TTS_Voices(OpenAI_Voice, GooeyEnum):
 
     ballad = OpenAI_Voice(label="Ballad (GPT-4o only)", voice_id="ballad")
     verse = OpenAI_Voice(label="Verse (GPT-4o only)", voice_id="verse")
+    marin = OpenAI_Voice(label="Marin (GPT-4o only)", voice_id="marin")
+    cedar = OpenAI_Voice(label="Cedar (GPT-4o only)", voice_id="cedar")
+
+    @classmethod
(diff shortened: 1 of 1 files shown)


devxpy (commented):
@milovate Small refactor suggestion — instead of `_gpt4o_supported_voices()` + `exclude=` in `openai_tts_selector`, consider a `supported_voices(model_name)` classmethod on `OpenAI_TTS_Voices` and build the selectbox from that list directly.

```python
@classmethod
def supported_voices(cls, model_name: str | None = None):
    model = OpenAI_TTS_Models.get(model_name)
    match model:
        case OpenAI_TTS_Models.gpt_4_o_mini | None:
            return list(cls)
        case _:
            return list(set(cls) - {cls.ballad, cls.verse, cls.marin, cls.cedar})
```

```python
def openai_tts_selector():
    model_name = gui.session_state.get("openai_tts_model")
    voices = OpenAI_TTS_Voices.supported_voices(model_name)
    options = {voice.name: voice.label for voice in voices}
    gui.selectbox(
        label="###### OpenAI Voice Name",
        key="openai_voice_name",
        options=list(options),
        format_func=options.__getitem__,
    )
```

Benefits:
- positive API ("what voices are supported?") instead of inverted exclude logic
- `None` grouped with `gpt_4_o_mini` fixes first-render hiding GPT-4o voices before model is set in settings
- reusable later in `TextToSpeech.run()` for API validation

devxpy (approved):

PR #1030, opened 2026-07-16
feat: Add tag filter to explore search

anish-work:
- Add `tag` field to SearchFilters and filter published runs by tag membership (case-insensitive, via subquery)
- Route tag pills/links (explore suggestions, home industry tiles, saved-workflow and header tags) through `tag=` instead of full-text `search=`, so tag pages use the featured/example_priority ordering rather than text rank
- Highlight the active tag pill with a clear (x) icon that removes the filter, and show the sort dropdown when a tag is selected
- Weight tag matches above title in the search vector (tags A, title B, workspace/author C, notes D)

### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

### DIFF
--- a/widgets/workflow_search.py
+++ b/widgets/workflow_search.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import html
 import re
 import typing
 
@@ -99,10 +100,13 @@ class SearchFilters(BaseModel):
     search: str = ""
     workspace: str = ""
     workflow: str = ""
+    tag: str = ""
     sort: str = ""
 
     def __bool__(self):
-        return bool(self.search or self.workspace or self.workflow or self.sort)
+        return bool(
+            self.search or self.workspace or self.workflow or self.tag or self.sort
+        )
 
     @field_validator("sort", "workflow", mode="before")
     @classmethod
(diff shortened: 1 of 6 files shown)


milovate (commented):
Added support for multi-select tag filtering, including intersection-based results and updated tag-chip UI states. Feel free to change or adjust anything as you see fit!

widgets/workflow_search.py
```diff
@@ -93,25 +93,42 @@ def html_icon_label(self) -> str:
...
+    def get_query_params(self) -> dict[str, str | list[str]]:
         return self.model_dump(exclude_defaults=True)
 
+    def toggle_tag(self, tag: str) -> SearchFilters:
+        tags = [
+            selected for selected in self.tag if selected.casefold() != tag.casefold()
+        ]
+        if len(tags) == len(self.tag) and len(tags) < MAX_TAG_FILTERS:
+            tags.append(tag)
+        return self.model_copy(update={"tag": tags})
```
devxpy, line 133: The `len(tags) == len(self.tag)` trick is an indirect way of asking "was it already selected?" — a more obvious version:

```suggestion
    def toggle_tag(self, tag: str) -> SearchFilters:
        tags = list(self.tag)
        for selected in tags:
            if selected.casefold() == tag.casefold():
                tags.remove(selected)
                break
        else:
            if len(tags) < MAX_TAG_FILTERS:
                tags.append(tag)
        return self.model_copy(update={"tag": tags})
```

devxpy (approved):

PR #1038, opened 2026-07-29
feat: enable editing controller-managed chat messages

anish-work:
### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate


devxpy:
**Refactor suggestion:** the `edit_sr` path in `chat_widget_input_to_request_body` shares nothing with the rest of the function — it ignores `state`, skips the button/location/history logic, and always returns `None` for `message_thread`. Meanwhile both callers (`VideoBots.on_send`, `gooey_builder_send_message`) duplicate the `edit_run_url` → `url_to_runs(...)` extraction before passing `edit_sr` in.

Suggested cleanup: drop the `edit_sr` parameter, have `chat_widget_input_to_request_body` read `edit_run_url` from `input_data`, and move the `url_to_runs` call into `_build_chat_widget_edit_request_body` (lazy import, matching the existing `handle_location_msg` pattern). Both call sites then collapse to a single unconditional call, and the now-unused `url_to_runs` import in `VideoBots.py` goes away.

Behavior is identical, including the pre-existing nuance that a malformed `edit_run_url` raises from `url_to_runs` inside the handler.

gooey-gui/app/components/GooeyBuilderInlineEmbed.tsx
```diff
@@ -66,16 +59,26 @@ export function GooeyBuilderInlineEmbed(
           let url = new URL(redirectUrl);
           ctx.current.navigate(url.pathname + url.search);
         },
+        editQuery: (_messageId: string, input_data: any) => {
```
devxpy, line 62: this should pass the stored `web_url` not _messageId

devxpy:
would work better with https://github.com/GooeyAI/gooey-server/pull/1049

anish-work:
Rasing a fresh PR

PR #1054, opened 2026-08-12
Gate document intelligence providers by configuration

milovate:
## Summary

- expose Azure Form Recognizer models only when `AZURE_FORM_RECOGNIZER_KEY` is configured
- expose Mistral OCR only when `MISTRAL_API_KEY` is configured
- show a linked configuration error and clear stale GUI model state when neither provider is available

## Why

Local deployments currently expose Mistral OCR without credentials and can retain the default Azure model even when no document-intelligence provider is configured. Users can then submit a model that is guaranteed to fail at runtime.

This keeps the existing request/API default intact for production compatibility while making GUI provider availability reflect the configured credentials.

## Validation

- `ruff check recipes/DocExtract.py`
- `git diff --check`

### DIFF
--- a/recipes/DocExtract.py
+++ b/recipes/DocExtract.py
@@ -607,7 +607,19 @@ def document_intelligence_settings(title: str, help: str | None = None):
     document_models = {}
     if settings.AZURE_FORM_RECOGNIZER_KEY:
         document_models |= azure_document_intelligence_models()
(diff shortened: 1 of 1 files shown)


devxpy:
Perhaps might be easy to skip ocr step? And just fallback to pandoc route

devxpy (approved):

PR #1068, opened 2026-08-22
feat: add GPT Transcribe ASR model

milovate:
## Summary\n\n- Add OpenAI's  as a selectable ASR model.\n- Reuse the existing OpenAI transcription request path and prompt support.\n- Keep the change limited to the ASR model registry and dispatch.\n\n## Testing\n\n- \n-  (1 skipped)


milovate:
Superseded by the clean follow-up PR #1069, which is based on current master and contains only the GPT Transcribe change.

PR #1086, opened 2026-09-02
fix: handle insufficient credits in builder

milovate:
- Builder credit failures bypassed the normal workflow error surface, so the sidebar and /new flow could lose retry actions, show a failed assistant placeholder, or drop existing conversation history.

- Reuse the shared credit renderer with a Builder-specific retry, preserve prior messages, reject cross-user retries, and keep the Builder header and embed bounded without changing other embed surfaces.

### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate

### DIFF
--- a/daras_ai_v2/gooey_builder.py
+++ b/daras_ai_v2/gooey_builder.py
@@ -94,9 +94,7 @@ def render_standalone_gooey_builder(
 
     handle_gooey_builder_redirect(builder_sr)
     builder_run_url = builder_sr.get_app_url()
-    messages = get_chat_widget_messages(
-        builder_sr.to_dict(), web_url=builder_run_url
-    )
+    messages = get_chat_widget_messages(builder_sr.to_dict(), web_url=builder_run_url)
     if builder_sr.error_type == exceptions.InsufficientCredits.__name__:
         render_gooey_builder_insufficient_credits(
             request=request,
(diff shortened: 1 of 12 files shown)


daras_ai_v2/gooey_builder.py
```diff
@@ -45,7 +45,11 @@ def render_gooey_builder(
 
     builder_sr = page.current_sr.parent_builder_saved_run
     handle_gooey_builder_redirect(builder_sr)
-
+    workflow_state = {
+        field_name: gui.session_state[field_name]
+        for field_name in page.fields_to_save()
```
devxpy: only needs page.RequestModel.model_fields i think?

widgets/errors.py
```diff
@@ -66,6 +67,8 @@ def insufficient_credits_error(error_params: dict):
     if gui.session_state.pop(RERUN_KEY, None):
         if rerun_workspace:
             set_current_workspace(request.session, rerun_workspace.id)
+        if on_rerun:
```
devxpy: Hmm, callbacks are generally an architectural nightmare to manage. Given that InsufficientCredits is a react component, perhaps what we can do is map the rerun button to the api call that we make in https://github.com/GooeyAI/gooey-server/blob/52487a126802bd3855557b4d86ff5a298efd25fe/gooey-gui/app/components/GooeyBuilderInlineEmbed.tsx#L75 ? if it works its worth trying i think.

milovate: agreed. The Builder now uses its own rerun key and sends the retry through submit_gooey_builder_message, which is the same server path used by the inline Builder.
This also fixed the Layout V2 bug. Layout V2 now converts the old -submit-workflow signal into a RunIntent, so the retry starts after the page reruns. I added a test for this case. Fixed in 7b36146e8.

devxpy (approved):

PR #1087, opened 2026-09-03
fix: switch agent panes without refetching

milovate:
- Agent editor tabs previously lived in Python session state, so selecting Knowledge, Tools, Settings, or Debug submitted the form and rebuilt the render tree over the network.

- Render every pane in the initial response and keep only the active pane ID in workflow-scoped React state, preserving mounted form content while tab clicks change visibility locally.

### Q/A checklist

<details>
<p>

```bash
```

```bash
```

```bash

```

```python
def my_function():
    import pandas as pd
    ...
```

</p>
</details>

### Legal Boilerplate


recipes/VideoBots_v2.py
```diff
@@ -375,37 +364,30 @@ def _render_about_meta_card(self, *, icon: str, label: str, pane: ConfigPane):
...
+        with gui.model_component(RecipeWorkspacePanesProps(panes=panes)):
+            with gui.div():
+                self._render_llm_instructions_pane()
+            with gui.div():
+                self._render_knowledge_pane()
+            # The Variables editor already renders beside the prompt with the same keys.
+            with gui.div():
+                self._render_functions()
+            with gui.div():
+                self._render_settings_pane()
```
anish-work, line 378: This is the crash. `_render_settings_pane` ends with `self.render_settings()` (VideoBots_v2.py:541) — that is v1's *entire* settings tab, and it contains an `if documents:` knowledge block that `_render_knowledge_pane` re-implements line for line. With one `render_pane()` call only one of the two ever ran; now both do.

Everything in that block therefore renders twice:

```
task_instructions            citation_style_selector      use_url_shortener
cache_knowledge_widget       doc_extract_selector         query_instructions_widget
keyword_instructions_widget  doc_search_advanced_settings
```

Two `<select name="citation_style">` in one tree is fatal because `applyFormDataTransforms` (`gooey-gui/app/renderer.tsx`) transforms per **node** and writes the result back into the shared `formData`, and `parseJSON` is not idempotent — the second node re-parses the first node's output:

```bash
node -e 'let f={citation_style:JSON.stringify("number_plaintext")}; const p=v=>JSON.parse(String(v)); f.citation_style=p(f.citation_style); f.citation_style=p(f.citation_style)'
```

That reproduces the error verbatim. Only bare-word values throw; `5`, `null` and `true` survive a double parse, which is why the other duplicated selects don't blow up — and why this looks intermittent rather than universal. Both duplicated blocks are gated on `documents`, so only knowledge-base agents hit it.

Suggested fix, which removes the cause rather than the symptom: extract v1's `if documents:` block into its own method on `VideoBotsPage` (e.g. `render_knowledge_settings()`) and have v1's `render_settings()` call it, so v1 is unchanged. Then `_render_knowledge_pane` calls that method instead of keeping its own copy, and `_render_settings_pane` calls only the remainder (glossary + `language_model_settings`). That also ends the drift between two hand-maintained copies of the same block.

daras_ai_v2/base_v2.py
```diff
@@ -707,39 +706,6 @@ def _top_bar_publish_label(self) -> str:
...
 
-    def _render_pane_strip(
-        self, panes: list[PaneSpec], *, key: str
-    ) -> typing.Callable[[], None]:
-        """Render an in-page strip of panes and return the one to draw.
-
-        Panes are panels within one view, so they have no url - an ordered list of specs and
-        a session-state key, which carries `PaneSpec.id` rather than the label.
-
-        Only the active pane renders. Widgets on the others keep their values because the
```
anish-work, line 718: Worth calling out that this guarantee — "Only the active pane renders" — is what was hiding the duplication above: whichever pane was active rendered its copy of the knowledge settings and the other never ran.

With it gone, every shared widget key now lives in one tree at once, so it is worth auditing for other duplicates beyond the knowledge block. The comment added at `VideoBots_v2.py:384` notes the Variables editor already renders beside the prompt with the same keys, so there is at least one more known pair.

anish-work (commented):
Heads up: this breaks the agent editor on `test` for any agent that has a knowledge base. Reported on load of `/agent/farmerchat-ag-advisory-agent-nuwsqmzp/` — the page renders `Uh oh … SyntaxError: Unexpected token 'm', "number_plaintext" is not valid JSON`. `/usage` on the same app is fine, because it has no editor.

The duplication that causes it is pre-existing; rendering every pane in one pass is what makes both copies live in the same render tree. Details inline.

---

Separately, and non-blocking (the file is not in this PR): now that panes render together and duplicate widget keys are legitimate — the comment added at `VideoBots_v2.py:384` notes the Variables editor already shares keys with the prompt — `applyFormDataTransforms` in `gooey-gui/app/renderer.tsx` should transform each form **name** once rather than each node. As it stands, any duplicated `select` or `file` input can corrupt form state or take the page down, which is what happens here. Threading a `Set` of already-transformed names through the recursion is enough. Happy to hand over that patch with tests; it is the guard rather than the fix, so it may belong in its own PR.

recipes/VideoBots_v2.py
milovate: Thanks—confirmed. The duplicate render path was the cause of the crash.
In this case, the duplicated block was also spec drift from v1. The v1 render_settings() implementation is the intended owner of the knowledge configuration, while the v2 Knowledge pane only needs to own document uploads.
I therefore removed the copied knowledge settings from _render_knowledge_pane() instead of extracting or restructuring the v1 implementation. _render_settings_pane() continues to call the existing v1 render_settings(), leaving v1 behavior unchanged and ensuring each form field is rendered once.
This removes the duplicate citation_style nodes and the resulting double parseJSON() failure for knowledge-base agents.
I also audited the Variables path. Variables renders once in the prompt dialog, while the Tools pane renders functions only. I clarified the nearby comment which might have poisoned the review.

daras_ai_v2/base_v2.py
milovate: Agreed. I audited the other panes; Variables renders only once in its dialog, while Tools renders functions only. I clarified the misleading comment and removed the confirmed Knowledge duplication. The generic form-transform safeguard can be handled separately

devxpy (approved):

PR #1089, opened 2026-09-05
feat: stream Fireworks chat completions

milovate:
## Summary

- Enable true incremental streaming for Fireworks chat completions when Gooey callers pass `stream=True`.
- Configure the existing OpenAI SDK client with the Fireworks API key and OpenAI-compatible base URL.
- Reuse Gooey existing cumulative stream accumulator for text, tool-call fragments, finish reasons, adaptive chunk thresholds, and usage recording.
- Preserve the current non-streaming request fields, return shape, structured-output gating, and cost accounting.

## Why this approach

Fireworks exposes an OpenAI-compatible Chat Completions endpoint, and this repository already depends on the OpenAI SDK and uses its typed stream objects. A provider-local adapter therefore keeps protocol handling at the transport boundary without changing callers, the frontend streaming contract, or unrelated providers.

This is more maintainable than routing Fireworks through the full OpenAI provider implementation because that path also contains OpenAI-specific reasoning-model and token-budget behavior. Keeping the adapter inside `_run_fireworks_chat` preserves Fireworks-specific semantics while sharing only the protocol-compatible client and accumulator.

It is also safer than parsing Fireworks SSE manually. The SDK owns SSE framing, blank events, UTF-8 decoding, JSON validation, `data: [DONE]`, HTTP errors, and response cleanup. Gooey only handles application-level `ChatCompletionChunk` objects through the same accumulator already used for OpenAI streaming. A manual `requests.iter_lines()` implementation would duplicate those protocol concerns and create additional edge cases around keepalives, usage-only events, fragmented tool arguments, cancellation, and leaked connections.

## Implementation

- Propagate `stream` and the existing adaptive chunk-size settings through the Fireworks provider boundary.
- Send `stream_options.include_usage=true` so final token accounting is explicit.
- Pass SDK stream chunks to `_stream_openai_chunked`.
- Use the existing OpenAI SDK retry predicate for SDK-native transport errors.
- Continue normalizing non-streaming SDK responses into the existing Fireworks message list.

## Validation

- Focused Fireworks adapter suite: 6 passed.
- Adjacent language-model, token-counter, and thinking-output suites: 18 passed.
- Ruff check and formatting verification passed.
- Live Fireworks smoke test produced three cumulative snapshots, a terminal `finish_reason=stop`, and exactly one prompt plus one completion usage record.

## Scope

This PR is intentionally isolated: one commit, one changed file, and no insufficient-credits or frontend changes.

The account live Fireworks catalog did not contain the model IDs currently seeded in `scripts/init_llm_models.py`. The adapter was validated with `accounts/fireworks/models/deepseek-v4-flash-0731`; model-catalog maintenance is a separate deployment follow-up.

### DIFF
--- a/daras_ai_v2/language_model.py
+++ b/daras_ai_v2/language_model.py
@@ -61,7 +61,6 @@ EMBEDDING_MODEL_MAX_TOKENS = 8191
 SUPERSCRIPT = str.maketrans("0123456789", "⁰¹²³⁴⁵⁶⁷⁸⁹")
 
 AZURE_OPENAI_MODEL_PREFIX = "openai-"
-FIREWORKS_OPENAI_BASE_URL = "https://api.fireworks.ai/inference/v1"
 
 
 class _ReasoningEffort(typing.NamedTuple):
@@ -403,8 +402,8 @@ def _run_chat_model(
                 response_format_type=response_format_type,
             )
         case ModelProvider.fireworks:
-            return _run_fireworks_chat(
-                model=model.model_id,
+            return run_openai_chat(
+                model=model,
                 avoid_repetition=avoid_repetition,
                 max_tokens=max_tokens,
                 messages=messages,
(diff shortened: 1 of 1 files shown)


milovate:
@codex

devxpy (approved):

PR #1090, opened 2026-09-08
feat: quote the interrupting message when a streamed WhatsApp reply is replaced

milovate:
## Summary

Follow-up to #1062. When a WhatsApp user sends a new message while a streamed reply is still going out, the active run is cancelled and its inputs are merged into a replacement run. The replacement reply now quotes the interrupting message on its first chunk, so the user can see which message the answer picks up from.

- `_cancel_active_run_and_merge_inputs` sets the quote target to the interrupting message id, but only when streaming is enabled and the cancelled run had already produced output text. If nothing was shown to the user yet, nothing is quoted.
- `WhatsappBot` threads that target through `send_msg_to` and `send_wa_msgs_raw` as WhatsApp's `context.message_id`. Only the first real chunk quotes. Button-only carriers such as feedback buttons or a location request never quote and do not consume the target.
- The target is cleared after the first reply chunk is sent, even if WhatsApp returns no message id. A failed send keeps it for the retry.

## Test plan

- [x] `tests/test_whatsapp_replies.py` (23 tests): first-chunk-only quoting for long, streamed, media, and button replies; carriers neither quote nor consume the target; failed sends keep it; the cancel path quotes only for streamed output and leaves completed, failed, cancelled, and missing runs alone.
- [x] `bots/tests.py` partial-reply tests still pass.
- [x] Manual: send a message to a streaming WhatsApp bot, send another mid-reply, confirm the next reply quotes the second message.

### DIFF
--- a/daras_ai_v2/bots.py
+++ b/daras_ai_v2/bots.py
@@ -797,6 +797,9 @@ def _cancel_active_run_and_merge_inputs(
     last_run.is_cancelled = True
     last_run.save(update_fields=["is_cancelled", "updated_at"])
     last_run.refresh_from_db()
+    # if the user already saw part of the old reply, quote the message that interrupted it
+    if bot.streaming_enabled and any(last_run.state.get("output_text") or []):
+        bot.reply_to_msg_id = bot.user_msg_id
     return _merge_run_inputs(last_run.state, request_body)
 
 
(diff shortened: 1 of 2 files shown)


milovate:
@codex

PR #1097, opened 2026-09-15
fix: media preview lightbox — Safari/iOS bugs, custom video controls, dialog polish

samarthgulati:
Follow-up to #1096 (already merged). That PR shipped with a few bugs that only
showed up after merging to gooey-ui-prod — this addresses those, plus a round
of polish based on comparing against gooey-web-widget's MediaPreview.tsx, and
the bot review findings on this PR itself.

## Fixes carried over from prod bugs
- Expand buttons were submitting the page's root form on click (missing `type="button"`)
- Preview dialog didn't trap/restore keyboard focus
- Safari's AirPlay icon overlapped the expand button on the inline thumbnail
- Preview dialog backdrop left a gap at the notch/Dynamic Island on iOS

## New: custom inline video controls
Native video controls on the inline thumbnail were the root cause of the
AirPlay-icon and corner-rounding bugs above — replaced entirely with a
custom overlay instead of chasing more browser-specific control-bar fixes:
- Autoplay/muted/loop preview, with play/pause (center), mute/unmute
  (top-right — deliberately claiming the corner Safari's own AirPlay icon
  wants, rather than contesting it there again), expand (top-left, matching
  the image expand button's position for consistency)
- All three fade in/out together based on activity (mouse movement, a tap,
  direct hover/focus on a button), 1.5s idle timeout, on both desktop and
  touch — a persistently visible pause button while a video is actively
  playing doesn't make sense on either, especially for generated clips that
  are often only 3-4s long. A reveal-tap on touch only reveals, it doesn't
  also act, so the same tap can't accidentally trigger audio.
- Pauses automatically when scrolled out of view (`IntersectionObserver`),
  resumes when scrolled back — independent of an explicit user pause, which
  sticks regardless of visibility.
- Pauses the inline copy while the dialog is open, so its own autoplaying
  video doesn't play twice at once, then restores whatever state the
  thumbnail was actually in before.
- Expand icon mirrored (`scaleX(-1)`) to match iOS's native expand glyph
  direction, since FontAwesome only ships the opposite diagonal.

## Dialog polish
- Action bar (Download/Copy link/Close) now floats via `position: fixed` on
  the viewport corner instead of `position: absolute` nested in the media
  surface — it no longer overlaps the video content or native controls.
- Download and Copy link have text labels now; Close stays icon-only.
- Media now sits in a rounded dark surface panel (matching
  gooey-web-widget's MediaPreview) instead of floating directly on the
  blurred backdrop.

## New: animated thumbnail → lightbox transition
Opening/closing the dialog now morphs the media card's position, size, and
border-radius between the thumbnail and the dialog via the native View
Transitions API, instead of the dialog just appearing/disappearing
instantly. No new dependency. Falls back to today's plain instant behavior
where unsupported (older Safari, Firefox) or when the user has asked for
reduced motion — not a regression either way. For video this crossfades a
snapshot rather than keeping playback visually continuous through the morph
(the API doesn't reparent a live element); accepted as matching how most
sites with this pattern behave.

## Bot review fixes (this PR)
- `handleDownload` now rejects unsuccessful HTTP responses instead of
  downloading a 404/403 page as if it were the media file
- Dialog focus-management effect no longer depends on the (frequently
  recreated) `onClose` callback identity, which was tearing it down and
  re-running it — bouncing focus out to the trigger and back — on every
  state change while the dialog was open
- Download blob URL revoke delayed to avoid racing Safari/iOS's queued
  download navigation
- Video play-sync effect now re-evaluates when `src` changes, so a recipe
  rerun swapping the video URL on the same component instance doesn't leave
  actual playback out of sync with pause state
- Download's fallback `window.open()` moved out of the async catch block
  and into a placeholder opened synchronously at click time, so it can't
  get silently blocked once Safari/iOS's transient activation window has
  expired

## Test plan
- [ ] Generate a video on /video, verify inline hover/tap controls, mute,
      pause, scroll-away auto-pause
- [ ] Open the lightbox, verify Download/Copy link/Close don't overlap the
      video, and the thumbnail→lightbox animation looks right (and its
      absence is fine on unsupported browsers)
- [ ] Try Download on Safari/iOS specifically
- [ ] Safari on iOS: check notch/Dynamic Island area, AirPlay icon, rounded
      corners


gooey-gui/app/components/MediaTags.tsx
```diff
@@ -201,41 +412,234 @@ const expandButtonStyle: React.CSSProperties = {
...
+  top: "50%",
+  left: "50%",
+  transform: "translate(-50%, -50%)",
+  width: 44,
+  height: 44,
+  fontSize: 16,
+  zIndex: 2,
+};
+
+const mediaDialogSurfaceStyle: React.CSSProperties = {
```
anish-work: The file defines nine module-level `React.CSSProperties` constants spanning ~115 lines, while the same component's visibility, radius, and margin rules live in `app.css`. Styling for one component is split across two mechanisms with no rule for which goes where.

can we make MediaTag.css and define all the css there and not use any inline styles until really necessary
Also we want to keep app.css is for global tokens

@samarthgulati

gooey-gui/app/components/MediaTags.tsx
```diff
@@ -47,22 +85,28 @@ export function GooeyImg({
       {clickable ? (
         <div
           className="gui-media-preview-wrap"
-          style={{ position: "relative", maxWidth: 450 }}
+          style={{
+            position: "relative",
+            maxWidth: 450,
```
anish-work: We already define this in app.css, do we need this on the wrapper?
and we can use position-relative from bootstrap and not use inline styles

gooey-gui/app/components/MediaTags.tsx
```diff
@@ -138,33 +264,124 @@ export function GooeyVideo({
...
-            onClick={() => setDialogOpen(true)}
-            style={expandButtonStyle}
+            // isPlayable: governed only by the timer-driven group below,
+            // same as mute/play-pause - .gui-media-expand-btn's plain
+            // :hover rule fires independent of that timer (it doesn't care
+            // whether the mouse has actually moved recently), so carrying
+            // both classes left this the only button still visible once the
+            // others faded out from under a motionless cursor.
+            className={
+              isPlayable ? "gui-video-overlay-btn" : "gui-media-expand-btn"
```
anish-work: use clsx package for dynamic classes

gooey-gui/app/components/MediaTags.tsx
```diff
@@ -138,33 +264,124 @@ export function GooeyVideo({
...
+          className={
+            "gui-media-preview-wrap" +
+            (controlsVisible ? " gui-video-controls-visible" : "")
+          }
+          style={{
+            position: "relative",
+            maxWidth: 450,
+            viewTransitionName: dialogOpen ? undefined : mediaTransitionName,
+          }}
+          onMouseMove={showControls}
```
anish-work, line 307: `onMouseMove={showControls}` runs `setControlsVisible(true)` plus a `clearTimeout`/`setTimeout` pair on every mousemove event — dozens per second while the cursor crosses the video.

Result:
Sweeping the mouse across a video fires ~60 events/sec, each allocating and discarding a timer. React bails out of the redundant `setControlsVisible(true)`, so the re-renders are free, but the timer churn is not — and on a page with several videos (CompareText2Img renders one per model) it multiplies. 

Suggestion:
Guard with `if (!controlsVisible) setControlsVisible(true)` and only reset the timer when it is actually due, or throttle the handler.
@samarthgulati

gooey-gui/app/components/MediaTags.tsx
```diff
@@ -201,41 +412,234 @@ const expandButtonStyle: React.CSSProperties = {
...
+  position: "fixed",
+  top: 16,
+  right: 16,
+  zIndex: 1000000,
+  display: "flex",
+  alignItems: "center",
+  gap: 8,
+};
+
+const actionPillButtonStyle: React.CSSProperties = {
```
anish-work: can we match the web-widget button styles since we are adding the buttons and also the backdrop styles to be consistent
https://github.com/GooeyAI/gooey-web-widget/blob/master/src/components/shared/Response/MediaPreview.tsx

 .gw-media-action-btn,
  .gw-media-icon-btn {
    display: inline-flex;
    align-items: center;
    gap: 6px;
    border: none;
    cursor: pointer;
    border-radius: 8px;
    padding: 8px 12px;
    font-size: 13px;
    font-weight: 600;
    transition: background 120ms ease, transform 120ms ease;
    color: #0b1021;
    background: #f1f5f9;
    border: 1px solid transparent;
  }

@samarthgulati

gooey-gui/app/components/MediaTags.tsx
```diff
@@ -201,41 +412,234 @@ const expandButtonStyle: React.CSSProperties = {
...
   cursor: "pointer",
+  flexShrink: 0,
+};
+
+// FontAwesome's up-right-and-down-left-from-center points along the
+// opposite diagonal from iOS's native expand/fullscreen glyph (up-left +
+// down-right) - there's no separate icon asset for that diagonal, so rotate
+// this one 90deg to match instead.
+const expandIconStyle: React.CSSProperties = {
+  // A horizontal mirror (rather than a 90deg rotation) turns the up-right/
```
anish-work: The comments are stale?

gooey-gui/app/components/MediaTags.tsx
```diff
@@ -19,6 +54,7 @@ export function GooeyImg({
 }) {
   const [previewIsValid, onError] = useImageValid(previewImg);
   const [dialogOpen, setDialogOpen] = useState(false);
+  const mediaTransitionName = mediaTransitionNameFor(src);
```
anish-work: This computes mediaTransitionNameFor(src) unconditionally, before the clickable check that excludes data: URIs. gui.image does emit data URIs for numpy arrays (common.py:287) — bounded by a 128px resize, so tens of KB, not megabytes. Still a global regex over a large string on every render, producing a value that's then thrown away. Move it below clickable.
@samarthgulati

gooey-gui/app/components/MediaTags.tsx
samarthgulati: Good catch - maxWidth:450 was redundant (.gui-img/.gui-video already cap at 450px, and the wrapper has no fixed width so it just shrinks to fit). Dropped it, and position:relative now comes from Bootstrap's position-relative utility instead of inline style. See 8da3b4cc.

gooey-gui/app/components/MediaTags.tsx
samarthgulati: Done, using clsx for all the dynamic classNames now. See 8da3b4cc.

gooey-gui/app/components/MediaTags.tsx
samarthgulati: Good catch, this was a real cost multiplying across a page with several videos. Throttled to actually reset the timer at most once per 200ms while already visible; the first reveal (from hidden) still runs immediately so it still feels instant. See 9c1197c2.

gooey-gui/app/components/MediaTags.tsx
samarthgulati: Matched .gw-media-action-btn/.gw-media-icon-btn's colors and radius for Download/Copy link/Close, and the backdrop's background+blur, to the widget's values. See 8da3b4cc.

gooey-gui/app/components/MediaTags.tsx
samarthgulati: Yes - the expand icon's outer comment still said "rotate this one 90deg" from an earlier round, but the code was already changed to scaleX(-1) per a later request and only the inner comment got updated to match. Fixed. See 8da3b4cc.

gooey-gui/app/components/MediaTags.tsx
samarthgulati: Good catch. Moved below clickable and made it conditional (computed only when clickable is true), so a non-clickable data: URI image doesn't run the regex over that string just to discard the result. See 8da3b4cc.

gooey-gui/app/components/MediaTags.tsx
samarthgulati: Moved everything into a new MediaTags.css, following the NavigationSidebar.css/RecipeTopBar.css convention. app.css now only has the global `img` reset left for this component's concern. Only view-transition-name stays inline - it's genuinely per-instance/dynamic. See 8da3b4cc.

anish-work (approved):

PR #1098, opened 2026-09-15
feat: run metadata in the v2 debug pane

milovate:
## Summary

- The Debug tab of the v2 editor now opens with a key/value block for the run: platform badge and masked end-user id for bot runs, the conversation title linked to its latest run, who ran it, which workspace was charged, and a link to the parent run.
- The "Timings" and "Prompt details" headings are removed; the sections are separated by rules.
- Fixes the metadata layout, which was stacking vertically because `gui.styled` stamps its class on each direct child rather than a wrapper, so the grid selector never matched.
- Pane styles live in `RecipeWorkspace.css` next to the other pane rules (following the `.v2-run-error` precedent) instead of a `gui.styled` string. Links in the metadata are always underlined, and the JSON viewer is pinned to the timeline's font size inside the pane.
- The editor panes' scroll region and the navigation sidebar now share one `.gooey-thin-scroll` utility in `app.css` instead of two identical copies.

## Not included

- A "Total time" row was tried and dropped: the only finish timestamp available is `updated_at`, which is `auto_now` and bumped by flagging, cancelling, and retention writes, so the value would be wrong after any later save. A real total needs a stored finished-at.
- The metadata block has no `can_view_usage`-style gate. It is admin-only today via `can_use_layout_v2`; that gate should be added before the v2 rollout widens.

## Test plan

- [ ] Open a copilot run in the v2 editor as an admin, switch to Debug, confirm the metadata renders as a two-column grid with labels on the left.
- [ ] Check a bot run shows the platform badge and masked user id, and a web run shows neither.
- [ ] Confirm the scrollbar in the editor panes matches the navigation sidebar.
- [ ] Confirm JSON in the Debug pane renders at the timeline's size.


daras_ai_v2/base_v2.py
```diff
@@ -1089,6 +1095,84 @@ def _usage_workspace(self) -> Workspace:
             return published_run.workspace
         return self.current_workspace
 
+    def render_debug_pane(self):
```
devxpy, line 1098: hmm, these would work better as a react component, perhaps extending RunTimeline

milovate: sure ill raise a separate PR to keep it clean

milovate:
Closing in favour of #1101, which reimplements the run metadata block as a `RunDebugInfo` React component fed by pydantic props instead of server-side `gui.*` markup.

This PR's commits were already merged into `feat/v2-layout-public-release` via 716a30c21; #1101 will be retargeted onto that branch and replaces the server-side pane there, so this one should not be merged to master.

PR #1102, opened 2026-09-16
fix: clear stale raw_tts_text between streamed chunks

milovate:
## Problem

Copilot TTS audio (and therefore lipsync video) was sometimes truncated to the first sentence or paragraph of the reply, while `output_text` showed the full reply. Reports have been inconsistent because the trigger depends on where the provider splits the stream.

## Cause

`output_translation_step` in `daras_ai_v2/harness.py` runs once per streamed chunk. It stripped HTML and whitespace from the chunk and wrote `raw_tts_text` only when the result differed from the display text, and never cleared it.

1. A mid-stream chunk ending in a newline (the model starting a new paragraph) differs after `.strip()`, so its partial text is stored.
2. The final chunk needs no stripping, compares equal, and leaves the stale partial value in place.
3. `tts_step` prefers `raw_tts_text` over `raw_output_text`, so only the partial text is spoken.

This is a regression from 5afbd2415 (May 6, 2026), which added the `.strip()`. Before that, the guard only fired when the model emitted real HTML mid-stream.

## Fix

Own the field on every pass: reset it to `None` when the stripped text matches the display text, otherwise assign it. The `None` contract for readers (`tts_step`, Slack `get-response-for-msg`, pricing) is unchanged.

## Verification

Ran the function twice on the same response object with the reply from the reported run: first a chunk ending after "paper cone.\n", then the full text. Before: `raw_tts_text` stayed at the first sentence. After: it is `None`, so TTS falls back to the full text. HTML button and `<think>` inputs still strip as before.


milovate:
@codex

devxpy (approved):

PR #1104, opened 2026-09-18
fix: bound upload metadata filenames

milovate:
Long image prompts are used as output filenames. The storage object name is already shortened, but upload registration saves the original filename into the indexed FileMetadata.name field. This caused issue 7455561066: a 3,224-byte B-tree entry exceeded PostgreSQL's 2,704-byte limit after image generation succeeded.

Reuse truncate_filename at metadata creation with a 255-character limit. This is a one-line production change, preserves Unicode and the filename's beginning and ending, and leaves names of 255 characters or fewer unchanged. The limit bounds the name to at most 1,020 UTF-8 bytes. No migration or new abstraction is needed. The behavior change applies to newly registered upload display names; existing metadata and full run prompts are unchanged.

Validation: before removing the test file from this PR, three local upload-path regression cases passed (short Unicode name, exactly 255 characters, and a long Unicode prompt). The long-name case failed with the production fix removed. Ruff lint, formatting, and git diff checks passed. Database and cloud operations were mocked; a live PostgreSQL integration test was not run. This PR contains only the one-line production fix.

https://dara-network.sentry.io/issues/7455561066/?project=4504338635423744&query=OperationalError&referrer=issue-stream&sort=freq


milovate:
@codex

devxpy (approved):

